PR #6, opened 2013-12-09
Update to allow better extension

allenxwang:
- Changed the pipeline configuration to keep track of HTTP state so that Observer.onCompleted() will be called appropriately.
- HttpProtocolHandler.configure() will be called after channel is established, instead of in the middle of channel read.
- Remove the code from ObservableHttpClient to dynamically install SSE decoder. Instead, SSE handling is done by specifying SSEHandler at execute() call.

### DIFF
--- a/src/main/java/rx/netty/protocol/http/ObservableHttpClient.java
+++ b/src/main/java/rx/netty/protocol/http/ObservableHttpClient.java
@@ -24,17 +24,20 @@ import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.SocketChannel;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.handler.codec.MessageToMessageDecoder;
 import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpClientCodec;
+import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpObject;
 import io.netty.handler.codec.http.HttpObjectAggregator;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpResponse;
+import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.util.AttributeKey;
 import io.netty.util.concurrent.EventExecutor;
 import io.netty.util.concurrent.Future;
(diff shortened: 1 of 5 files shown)


benjchristensen:
Thank you Allen!

src/main/java/rx/netty/protocol/http/ObservableHttpClient.java
```diff
@@ -166,42 +174,44 @@ private void tryUpdateUserAgent(ValidatedFullHttpRequest request) {
...
-                final ObservableHttpResponse<T> httpResponse = new ObservableHttpResponse<T>(response, PublishSubject.<T> create());
                 observer.onNext(httpResponse);
-                ctx.channel().pipeline().addLast("content-handler", new HttpMessageObserver<T>(observer, httpResponse));
-
-                handler.configure(ctx.channel().pipeline());
+            }
+            if (msg instanceof HttpContent) {
+                ((HttpContent) msg).content().retain();
+            }
+            ctx.fireChannelRead(msg);
```
g9yuayon, line 202: This line will get executed even _after_ this handler (HttpObservableTracker) is removed by SSEHandler. The result is a silent onError() call due to ClassCastException. This in theory should not happen at all. We need to find the root cause and fix it.

allenxwang, line 202: I think what might happen is that HttpClientCodec has taken the first chunk of bytes and decoded into two parts: 1) HttpResponse, 2) first HttpContent. While HttpClientCodec is removed after the HttpResponse is delivered and processed, Netty seems to understand that it yet needs to deliver the second piece, which is the first HttpContent, to the same pipeline before it is rearranged. And with the original code which uses MessageToMessageDecoder, even though it does not throw any error, the first HttpContent may have been ignored.

What we can do is that for SSEHandler, it should assume that HttpContent may be delivered from the pipeline and install a necessary decoder that decodes from HttpObject to Message.

NiteshKant, line 202: Here is the problem:

Your SSEHandler code, is:

```
        pipeline.remove("http-codec");
        pipeline.remove(NAME);
        pipeline.replace("http-response-decoder", "http-sse-handler", new ServerSentEventDecoder());
```

The pipeline before executing the above code, looks like this:

http-codec -> http-response-decoder -> sse-handler -> content-handler

What you want to transform this to, is:

http-sse-handler -> content-handler

However, the first thing SSEHandler does is to remove http-codec from the pipeline. Since, the http-codec is removed, no pipeline changes made after this removal (removal of tracker) are visible to the codec itself. As Allen noted, the ClientCodec has two messages HttpResponse & HttpContent. The handling of HttpResponse changes the pipeline but the changes are not visible to the ClientCodec (as I mentioned before) & hence it still thinks, it should call the tracker next, which it does while publishing HttpContent. The relevant code can be found here:

https://github.com/netty/netty/blob/4.0/codec/src/main/java/io/netty/handler/codec/ByteToMessageDecoder.java#L153

The two callbacks HttpResponse & HttpContent are fired in that loop.

The correct way of modifying the pipeline would be to remove the "http-codec" at the end. If you change the code to:

```
        pipeline.remove(NAME);
        pipeline.replace("http-response-decoder", "http-sse-handler", new ServerSentEventDecoder());
        pipeline.remove("http-codec");
```

it will work as expected.

PR #19, opened 2014-01-28
Refactoring

NiteshKant:
Following were the major design goals in this refactoring:

1) Modify java packages to better indicate what they contain.
2) Have a common client & server implementation that can be used by all connection oriented protocols.
3) Simplify netty's pipeline to have a thin adapter to convert netty's messages to Rx Observables.
4) Remove any custom callbacks which can be obtained by having a handler in the netty pipeline.
5) Model all protocol handling as netty's pipeline handlers.

TODO:

1) Fix groovy examples.
2) Http request execution should be cancelled as much as possible when the subscription is unsubscribed.


g9yuayon:
I'll just merge the pull request so I can start using the updated API in Mantis.

PR #49, opened 2014-02-19
Renaming SSEEvent to ServerSentEvent

NiteshKant:
Also improving the SSE client & server example to remove unnecessary netty pipeline configurations.


benjchristensen:
> Renaming SSEEvent to ServerSentEvent

Much better :-)

Issue #53, opened 2014-02-26
Reusable HTTPRequests?


NiteshKant:
[HttpRequest](https://github.com/Netflix/RxNetty/blob/master/src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java) provided by HttpClient contains a ContentSource which is a collection of Objects that would be sent as payload of the HTTP request sent by HttpClient.

In order for [Ribbon](https://github.com/Netflix/ribbon) or any other higher level framework to provide retries, it is imperative that it should be able to send the same request to multiple clients. However, with the current model, the same HttpRequest can not be reused as the ContentSource contains state of the iteration.

Assuming we want to support usecases like this, without forcing the users to clone the object before every use, there is the following we have to do:
- The framework overwrites any state changes it makes to headers while sending the request, eg: The host header.
- The content source set in the HttpRequest should provide a way to reset state. This can be achieved in two possible ways:
  - Provide a reset method on the ContentSource. This is too intrusive for content sources that do not need reuse and providing a different interface adds to confusion.
  - Return a new instance of ContentSource, every time a request is to be sent i.e. instead of returning the same contentSource object [here](https://github.com/Netflix/RxNetty/blob/master/src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java#L130) return a new instance every time. In order to achieve this, we would have to optionally provide a override method of withContentSource() that takes a ContentSourceFactory. In case, this factory is set, we would use this factory to create the new content source instance, else we will use the content source object as is.

@benjchristensen @g9yuayon @allenxwang Thoughts?

allenxwang:
It is not clear what the intention of the withContent_() APIs is when such methods are called multiple times. To me, the with_ methods gives a feeling that the final state of the HttpRequest is the union of those. But it seems that the real intention is that they should override each other. Also, the raw content and content are mutually exclusive. Should we reflect those points in the API design and make the intention clear to the user?

NiteshKant:
@allenxwang do you have some suggestions as to how these can be more intutive?

PR #57, opened 2014-02-28
Fix Issue #53

allenxwang:
Fix Issue #53.

Created new JUnit test for POST with content.

### DIFF
--- a/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
+++ b/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
@@ -91,18 +91,13 @@ public class ClientRequestResponseConverter extends ChannelDuplexHandler {
         Class<?> recievedMsgClass = msg.getClass();
 
         if (HttpRequest.class.isAssignableFrom(recievedMsgClass)) {
-            @SuppressWarnings("rawtypes")
-            HttpRequest rxRequest = (HttpRequest) msg;
+            HttpRequest<?> rxRequest = (HttpRequest<?>) msg;
             if (rxRequest.getHeaders().hasContent()) {
                 MultipleFutureListener allWritesListener = new MultipleFutureListener(promise);
                 allWritesListener.listen(ctx.write(rxRequest.getNettyRequest()));
-                ContentSource<?> contentSource = null;
-                if (rxRequest.getContentFactory() != null) {
-                    contentSource = rxRequest.getContentFactory().newContentSource();
-                } else {
-                    contentSource = rxRequest.getContentSource();
-                }
-                if (contentSource instanceof RawContentSource) {
+                ContentSource<?> contentSource;
+                if (rxRequest.hasRawContentSource()) {
+                    contentSource = rxRequest.getRawContentSource();
                     @SuppressWarnings("rawtypes")
                     RawContentSource<?> rawContentSource = (RawContentSource) contentSource;
                     while (rawContentSource.hasNext()) {
@@ -113,6 +108,7 @@ public class ClientRequestResponseConverter extends ChannelDuplexHandler {
                         allWritesListener.listen(ctx.write(byteBuf));
                     }
                 } else {
+                    contentSource = rxRequest.getContentSource();
                     while (contentSource.hasNext()) {
                         allWritesListener.listen(ctx.write(contentSource.next()));
                     }
--- a/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
+++ b/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
@@ -96,8 +96,15 @@ public class ClientRequestResponseConverter extends ChannelDuplexHandler {
             if (rxRequest.getHeaders().hasContent()) {
                 MultipleFutureListener allWritesListener = new MultipleFutureListener(promise);
                 allWritesListener.listen(ctx.write(rxRequest.getNettyRequest()));
-                if (rxRequest.hasRawContentSource()) {
-                    RawContentSource<?> rawContentSource = rxRequest.getRawContentSource();
+                ContentSource<?> contentSource = null;
+                if (rxRequest.getContentFactory() != null) {
+                    contentSource = rxRequest.getContentFactory().newContentSource();
+                } else {
+                    contentSource = rxRequest.getContentSource();
+                }
+                if (contentSource instanceof RawContentSource) {
+                    @SuppressWarnings("rawtypes")
+                    RawContentSource<?> rawContentSource = (RawContentSource) contentSource;
                     while (rawContentSource.hasNext()) {
                         @SuppressWarnings("rawtypes")
                         ContentTransformer transformer = rawContentSource.getTransformer();
@@ -106,8 +113,6 @@ public class ClientRequestResponseConverter extends ChannelDuplexHandler {
                         allWritesListener.listen(ctx.write(byteBuf));
                     }
                 } else {
-                    @SuppressWarnings("rawtypes")
-                    ContentSource contentSource = rxRequest.getContentSource();
                     while (contentSource.hasNext()) {
                         allWritesListener.listen(ctx.write(contentSource.next()));
                     }
--- a/src/main/java/io/reactivex/netty/protocol/http/client/ContentFactory.java
+++ b/src/main/java/io/reactivex/netty/protocol/http/client/ContentFactory.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2014 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.reactivex.netty.protocol.http.client;
+
+/**
+ * A {@link ContentSource} factory where its {@link ContentFactory#newContentSource()}
+ * method should always return a new {@link ContentSource} that starts from the 
+ * beginning.
+ *
+ * @author awang
+ *
+ * @param <T>
+ */
+public interface ContentFactory<T, R extends ContentSource<T>> {
+    public R newContentSource();
+}
--- a/src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
+++ b/src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
@@ -31,18 +31,43 @@ import java.util.concurrent.atomic.AtomicBoolean;
  * @author Nitesh Kant
  */
 public class HttpRequest<T> {
+    static class SimpleContentSourceFactory<T> implements ContentSourceFactory<T, ContentSource<T>> {
+        private ContentSource<T> source;
+        
+        SimpleContentSourceFactory(ContentSource<T> source) {
+            this.source = source;    
+        }
+        @Override
+        public ContentSource<T> newContentSource() {
+            return source;
+        }
+    }
+    
+    static class SimpleRawContentSourceFactory<T> implements ContentSourceFactory<T, RawContentSource<T>> {
+        private RawContentSource<T> source;
+        
+        SimpleRawContentSourceFactory(RawContentSource<T> source) {
+            this.source = source;    
+        }
+        @Override
+        public RawContentSource<T> newContentSource() {
+            return source;
+        }
+    }
 
     private final io.netty.handler.codec.http.HttpRequest nettyRequest;
     private final HttpRequestHeaders headers;
-    protected ContentSource<T> contentSource;
-    protected ContentFactory<T, ?> contentFactory;
+    protected ContentSourceFactory<T, ContentSource<T>> contentFactory;
+    protected ContentSourceFactory<?, RawContentSource<?>> rawContentFactory;
+    protected boolean userPassedInFactory = false;
+
     private AtomicBoolean contentSet = new AtomicBoolean();
 
     /*Visible for testing*/ HttpRequest(io.netty.handler.codec.http.HttpRequest nettyRequest) {
         this.nettyRequest = nettyRequest;
         headers = new HttpRequestHeaders(nettyRequest);
-        contentSource = null;
         contentFactory = null;
+        rawContentFactory = null;
     }
 
     public static HttpRequest<ByteBuf> createGet(String uri) {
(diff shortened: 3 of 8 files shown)


src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
```diff
@@ -96,8 +96,15 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
             if (rxRequest.getHeaders().hasContent()) {
                 MultipleFutureListener allWritesListener = new MultipleFutureListener(promise);
                 allWritesListener.listen(ctx.write(rxRequest.getNettyRequest()));
-                if (rxRequest.hasRawContentSource()) {
-                    RawContentSource<?> rawContentSource = rxRequest.getRawContentSource();
+                ContentSource<?> contentSource = null;
```
NiteshKant: null initialization is redundant.

src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
```diff
@@ -96,8 +96,15 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+                if (rxRequest.getContentFactory() != null) {
```
NiteshKant: Would it be better if we do not expose the factory from HttpRequest? Ideally this code must just do request.getContentSource() and then request.isRaw(contentSource) to handle raw of otherwise content source.

src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
```diff
@@ -96,8 +96,15 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+                    contentSource = rxRequest.getContentFactory().newContentSource();
+                } else {
+                    contentSource = rxRequest.getContentSource();
+                }
+                if (contentSource instanceof RawContentSource) {
+                    @SuppressWarnings("rawtypes")
+                    RawContentSource<?> rawContentSource = (RawContentSource) contentSource;
```
NiteshKant, line 107: You don't need the wildcard <?> here.

src/main/java/io/reactivex/netty/protocol/http/client/ContentFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * A {@link ContentSource} factory where its {@link ContentFactory#newContentSource()}
+ * method should always return a new {@link ContentSource} that starts from the 
+ * beginning.
+ *
+ * @author awang
+ *
+ * @param <T>
+ */
+public interface ContentFactory<T, R extends ContentSource<T>> {
+    public R newContentSource();
```
NiteshKant: keyword public is redundant for interface methods.

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
```diff
@@ -0,0 +1,115 @@
...
+        super(request.getNettyRequest());
+        if (request.contentSource != null) {
+            this.contentFactory = new RepeatableContentFactory<T>(request.contentSource);
+        } else if (request.contentFactory != null) {
+            this.contentFactory = request.contentFactory;
+        }
+    }
+}
+
+class RepeatableContentFactory<T> implements ContentFactory<T, ReplayingContentSource<T>> {
```
NiteshKant: I guess you wanted to make these classes as inner classes and not multiple top level classes in the same file, rite?

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
```diff
@@ -33,14 +34,15 @@
 
     private final io.netty.handler.codec.http.HttpRequest nettyRequest;
     private final HttpRequestHeaders headers;
-    private RawContentSource<?> rawContentSource;
-    private ContentSource<T> contentSource;
+    protected ContentSource<T> contentSource;
+    protected ContentFactory<T, ?> contentFactory;
+    private AtomicBoolean contentSet = new AtomicBoolean();
```
NiteshKant, line 64: Since HttpRequest is not required to be threadsafe, you can do with a boolean instead of an AtomicBoolean. OTOH, it should be a final, if its an AtomicBoolean

src/main/java/io/reactivex/netty/protocol/http/client/ContentFactory.java
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * A {@link ContentSource} factory where its {@link ContentFactory#newContentSource()}
+ * method should always return a new {@link ContentSource} that starts from the 
+ * beginning.
+ *
+ * @author awang
+ *
+ * @param <T>
+ */
+public interface ContentFactory<T, R extends ContentSource<T>> {
```
NiteshKant: Will ContentSourceFactory be more indicative of what it is?

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
```diff
@@ -0,0 +1,115 @@
...
+            return next;
+        }
+    }
+        
+    public void rewind() {
+        iteratorOfRecorded = recorded.iterator();
+    }
+}
+
+class RecordingRawContentSource<T> extends ReplayingContentSource<T> implements RawContentSource<T> {
```
NiteshKant: If you are a fan of consistent naming, it should be called ReplayingRawContentSource :)

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public ReplayingContentSource<T> newContentSource() {
+        recordingSource.rewind();
+        return recordingSource;
+    }
+}
+
+class ReplayingContentSource<T> implements ContentSource<T> {
+    private ContentSource<T> original;
+    private LinkedBlockingQueue<T> recorded = new LinkedBlockingQueue<T>();
```
NiteshKant: Since, this is not supposed to be threadsafe, you may just use a list.

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
```diff
@@ -0,0 +1,115 @@
...
+            this.contentFactory = new RepeatableContentFactory<T>(request.contentSource);
+        } else if (request.contentFactory != null) {
+            this.contentFactory = request.contentFactory;
+        }
+    }
+}
+
+class RepeatableContentFactory<T> implements ContentFactory<T, ReplayingContentSource<T>> {
+
+    private ReplayingContentSource<T> recordingSource;
```
NiteshKant: For the sake of consistency, this should be called replayingSource.

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+    
+    public ReplayingContentSource(ContentSource<T> original) {
+        this.original = original;
+        iteratorOfRecorded = recorded.iterator();
+    }
+    
+    @Override
+    public boolean hasNext() {
+        if (iteratorOfRecorded.hasNext()) {
```
NiteshKant: can be replaced by iteratorOfRecorded.hasNext() || original.hasNext()

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+        
+    public void rewind() {
+        iteratorOfRecorded = recorded.iterator();
+    }
+}
+
+class RecordingRawContentSource<T> extends ReplayingContentSource<T> implements RawContentSource<T> {
+    
+    private RawContentSource<T> original;
```
NiteshKant: Since, the only thing you need from the original source here is the transformer, you may just store that & it will be nice to make this variable final.

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
```diff
@@ -33,14 +34,15 @@
 
     private final io.netty.handler.codec.http.HttpRequest nettyRequest;
     private final HttpRequestHeaders headers;
-    private RawContentSource<?> rawContentSource;
-    private ContentSource<T> contentSource;
+    protected ContentSource<T> contentSource;
```
NiteshKant: Thinking about this more, I think it is better to only have a contentFactory and a rawContentFactory here and wrap all withContent() as a factory. That way you don't have to check for content already set & would make this code simpler.
It will be imperative to have both contentFactory & rawFactory as the rawFactory type would be a wild card &lt;?&gt; whereas ContentFactory will be &lt;T&gt;

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
```diff
@@ -33,14 +34,15 @@
...
+    protected ContentFactory<T, ?> contentFactory;
```
NiteshKant: The appropriate generic here is ContentFactory&lt;T, ? extends ContentSource&lt;T&gt;&gt;

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
```diff
@@ -130,12 +146,7 @@ public String getUri() {
         return contentSource;
     }
 
-    RawContentSource<?> getRawContentSource() {
-        return rawContentSource;
-    }
-
-    boolean hasRawContentSource() {
-        return null != rawContentSource;
+    final ContentFactory<T, ?> getContentFactory() {
```
NiteshKant: I think we should not expose the factory. getContentSource() should just do a factory.newSource()

NiteshKant:
Apologies, I had to merge my pull request, so you would have to merge.

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
allenxwang, line 64: I prefer the AtomicBoolean as it makes the check and set code clean.

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
allenxwang: Actually using a non concurrent list will fail the JUnit test `RepeatableRequestTest`. The reason is that even in the same thread, the content of the list may be altered after its iterator is created, causing ConcurrentModificationException thrown.

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
allenxwang: I thought we would need to check for content already set anyway, right? Otherwise it leads to unclear behavior to user when they can set multiple content with either withContent() or withContentFactory().

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
allenxwang: What's the difference? These are package private APIs anyway and more like implementation details.

allenxwang: OK, thinking about this again I realized that doing this will encapsulate the logic in one class and help our own APIs that uses HttpRequest does not have to repeat the same check whether it is a factory or source. Is that the intention?

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
allenxwang: `RepeatableContentHttpRequest` needs to access the content source of the wrapped request, and if there is no source but only factory it will not wrap it again with `RepeatableContentFactory`. So removing contentSource would make its logic complicated.

Another solution would be wrap all withContent() calls with withContentSource() or withRawContentSource(). And remove the access to factory, thus moving all the logic to the getContentSource() the only way to get the source.

src/main/java/io/reactivex/netty/protocol/http/client/RepeatableContentHttpRequest.java
NiteshKant: I see, then, the use of iterator is not correct. Looking at the code, it will be difficult for anyone to understand as to why this is a concurrent data structure when the class does not need to be threadsafe.
I guess all you need is an arraylist and the current index.

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
NiteshKant: So, what I am saying is that you will only have two instance variables, viz.,

ContentFactory<T, ContentSource<T>> contentFactory;
ContentFactory<?, RawContentSource<?>> rawContentFactory;

Now, methods withContent() will set the contentFactory as a ContentFactory implementation that just returns the same instance all the time. Methods withRawContent() will set the rawContentFactory. On top of it you will have methods to set the contentFactory and rawContentFactory.

With the above, you do not need to check if the content/factory is set or not as you will always overwrite the factory.

The other package-private methods getContentSource(), getRawContentSource() and hasRawContent() remains the same and so does the usage of these.

getContentSource() will return contentFactory.newSource()
getRawContentSource() will return rawFactory.newSource()
hasRawContent() will return null != rawFactory

Does this make sense?

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
NiteshKant: Yes thats the intention.

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
NiteshKant: Sorry I did not see your other note. Since contentSource is not final, I would say its safer to not make it protected. Instead, you can just call withContentFactory() from RepeatableContentHttpRequest

" thus moving all the logic to the getContentSource() the only way to get the source."
Yes thats what I was referring to.

src/main/java/io/reactivex/netty/protocol/http/client/HttpRequest.java
```diff
@@ -79,17 +106,46 @@
     }
 
     public HttpRequest<T> withContentSource(ContentSource<T> contentSource) {
-        this.contentSource = contentSource;
+        if (!contentSet.compareAndSet(false, true)) {
```
NiteshKant, line 109: Since, we are overwriting the content factory, why do we need this check? 
The con is that it limits overriding the content later, in case it is to be reused.

allenxwang, line 109: My intention is that only one method from withContent*() can called. User should create new Request object to set a different content source/factory.

NiteshKant, line 109: Sure but that constraint is not adding value to the code in terms of state
management, code clarity, etc. OTOH, it is making it difficult to reuse.
So, it seems to me that it is not required. Specially, since this is a
public API, I would recommend removing such a constraint.

allenxwang, line 109: Fair point. I have removed such checking in pull request

https://github.com/Netflix/RxNetty/pull/60

I will add more questions/comments on the issue itself.

PR #59, opened 2014-03-01
Removing compile time warnings and other redundant constructs.

NiteshKant:


allenxwang:
I am wondering if we should enforce coding style or if there is real benefit to be strict about this. For example, why "public" should be always omitted from interface and "this" should be omitted from instance variables?

NiteshKant:
I don't think it's about "always" omitting, its about omitting when it's
redundant. public keyword for interface methods is redundant and so is for
instance variables when there is no conflict.
The irritant about having them in the code is that a good IDE with default
code style shows them as warnings. Unless there is a reason for keeping it,
I guess it's good to remove them.

allenxwang:
I am using Eclipse and Intellij and neither of them report these as warnings.

Looking into other open source libraries, most of them have mixed style. Interestingly, `Observer` interface in RxJava not only has "public", but also "abstract".

I believe this is a subjective manner, as most of the stackoverflow discussions on such topics have called out. My personal feeling is that a little redundancy like this in the code does not  hurt. It may help readability in non IDE and text editors without syntax highlighting. 

I think a bigger question is whether we should enforce certain code style. If certain coding style is important to enforce, let's do it in build process. Otherwise, let's go easy on this.

Issue #74, opened 2014-03-20
Use PooledByteBufAllocator.DEFAULT by default as Bootstrap option


allenxwang:
According to the talk given by Norman Maurer, use PooledByteBufAllocator greatly reduces GC pressure.

PR #75, opened 2014-03-24
Fix issue #74

allenxwang:
- Also invokes `defaultChannelOptions()` in the builder constructor.
- For server builder, use `bootstrap.childOption()` instead of `bootstrap.option()`

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/client/AbstractClientBuilder.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/client/AbstractClientBuilder.java
@@ -16,6 +16,7 @@
 package io.reactivex.netty.client;
 
 import io.netty.bootstrap.Bootstrap;
+import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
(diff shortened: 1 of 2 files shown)


rx-netty/src/main/java/io/reactivex/netty/server/AbstractServerBuilder.java
```diff
@@ -69,12 +70,13 @@ public B channel(Class<ServerChannel> serverChannelClass) {
     }
 
     public <T> B channelOption(ChannelOption<T> option, T value) {
-        serverBootstrap.option(option, value);
+        serverBootstrap.childOption(option, value);
```
NiteshKant, line 73: Since, we have to have both parent and child socket options, we should add a new method childChannelOption() instead of changing the behavior for this method.

rx-netty/src/main/java/io/reactivex/netty/client/AbstractClientBuilder.java
```diff
@@ -42,6 +43,7 @@ protected AbstractClientBuilder(Bootstrap bootstrap, String host, int port) {
         this.bootstrap = bootstrap;
         serverInfo = new RxClientImpl.ServerInfo(host, port);
         clientConfig = RxClient.ClientConfig.DEFAULT_CONFIG;
+        defaultChannelOptions();
```
NiteshKant, line 46: The same needs to be done for the AbstractServerBuilder

allenxwang:
The change and code to address the review comments are in https://github.com/Netflix/RxNetty/pull/86

Issue #76, opened 2014-03-25
Channel pool should listen for channel close events


NiteshKant:
Currently the channel pool implementation does not have a way to react to close of underlying socket. It would depend on the next reuse of the connection or a background thread to detect such socket closures.
A scenario in which this limitation is highlighted is a ReadTimeoutHandler closing the connection after firing the timeout exception. In this case, the underlying connection will be in the idle pool till it is picked for reuse and a close is detected.

allenxwang:
If we put the channel back to the idle pool without closing it, the next time it is picked for reuse and the read timeout has not happened yet, it will cause problem as we are sending another request on the channel without getting the response from the first request, right? Does HTTP protocol allow sending two requests on the same socket without getting the response from the first?

allenxwang:
ConnectionLifecycleHandler remains in the pipeline after a channel is returned to the pool. Currently the channelUnregistered() method in this handler calls connection.close() which returns the channel to the pool. The pool will remove the channel if it is already closed and not reusable.

 Is channelUnregsitered() automatically called when channel is closed? If not, we just need to make connection.close() is called when channel is closed. Perhaps by overriding channelInactive()?

NiteshKant:
> > > If we put the channel back to the idle pool without closing it

May be I was not clear in my description. I did not mean to put the channel back into the reusable pool before it is closed. I meant that; this is what will happen, if we close the connection after the request processing is over i.e. when readtimeout occurs. This was just an example. The basic problem is that we are not listening to socket close events, which can happen without user interaction, eg: Channel closed by the server.

> > > ConnectionLifecycleHandler remains in the pipeline after a channel is returned to the pool. Currently the channelUnregistered() method in this handler calls connection.close() which returns the channel to the pool. The pool will remove the channel if it is already closed and not reusable.

In the case when we close the connection (fix for issue #21 ) after we are done, conn.close() in channelUnregistered() will be a no-op as ObservableConnection ignore multiple close on the same connection. So, this call will not do what you expect it to do i.e. have isReusable() return false in the pool implementation.

> > > Is channelUnregsitered() automatically called when channel is closed?

Channel unregistered is called when the channel is unregistered from the eventloop which happens after channelInactive() is invoked i.e. when the channel is closed.

> > > If not, we just need to make connection.close() is called when channel is closed. Perhaps by overriding channelInactive()?

[...]

Issue #21, opened 2014-01-28
HttpClient should cancel request write & response processing on unsubscription


NiteshKant:
Currently HttpClient cancels the connection subscription when the caller of HttpClient.submit() unsubscribe. This in turn closes the connection.
The relevant code is here: 

https://github.com/Netflix/RxNetty/blob/master/src/main/java/io/reactivex/netty/protocol/http/HttpClientImpl.java#L71

Ideally, it should do the following:

1) If write is not completed, cancel write, which if successful should close connection.
2) If the response is not completed, we wait for response to complete and then close the connection.

In the above statements, close the connection in presence of connection pooling (https://github.com/Netflix/RxNetty/issues/20) would signify connection returning to pool. In such a case, the connection should be in a sane state to re-use i.e. no read/write should be pending.

allenxwang:
I was under the impression that if the Observer is unsubscribed, the channel should be closed in current implementation. But it does not seem to be true. Here are the steps I observed during debugging:
- Request is successfully written to the pipeline and response is also received in the pipeline. onCompleted() method of the Observer that is subscribed to `Observable<HttpResponse>` is called.
- the above Observer is automatically unsubscribed
- Line `connectSubscription.unsubscribe();` inside HttpClientImpl.submit() is called. However, stepping into this call shows that the internal state is already "unsubscribed", so nothing really happened. Another observation is that the method `ObservableConnection.close()` is never invoked.

So it seems that the channel still remains open.

NiteshKant:
> > Line connectSubscription.unsubscribe(); inside HttpClientImpl.submit() is called. However, stepping into this call shows that the internal state is already "unsubscribed", so nothing really happened.

The unsubscribe from connectObservable does not close the connection as it just means, the subscriber does not want to get any new connections. The connection close is explicit, which does not seem to be happening currently in HttpClient. HttpClient should call close on the connection when the request processing is completed. 
I will fix this issue.

NiteshKant:
Now RxClient does cancel all pending writes on unsubscribe (or connection close) but it does not explicitly close a pooled connection if the response is not yet over.

NiteshKant:
This was fixed as part of #208 in release 0.3.15

PR #78, opened 2014-03-25
Fix for issue #21

NiteshKant:
1) Invoking close connection when the Observable returned from HttpClient.submit() is unsubscribed.
2) Fixed java examples, they were broken after the name change of HttpRequest/Response.
3) Modified a test case HttpClientTest.testReadtimeoutCloseConnection() to not check for deleted connections from the pool due to issue #76


allenxwang:
Can we also make compiling the Java example part of build? Otherwise we always have to remember updating these when API is changed.

Issue #89, opened 2014-04-04
Default Client and Server Should Use Same Netty EventLoops


benjchristensen:
Right now we create different event loops for both client and server. They should reuse the same as it is more efficient and right for most use cases.

benjchristensen:
Here is how it defaults right now:

```
Daemon Thread [rx-server-selector-1-1] (Running)    
Daemon Thread [rx-server-selector-1-2] (Running)    
Daemon Thread [rx-client-selector-3-1] (Running)    
Daemon Thread [rx-client-selector-4-1] (Running)    
```

There should not be "client" and "server" selectors, just "rx-netty-selector" threads, sized to the number of cores on the machine.

NiteshKant:
Created a pull request #92 for this. It will also fix the other issue where every RxClient instance was creating a new eventloop group.

benjchristensen:
Thank you!

PR #92, opened 2014-04-09
Fix for issue #89

NiteshKant:
Adds a capability to specify a provider for netty's eventloops for Rx clients and servers when they have not been specified explicitly.

By default, we will use the same eventloop for both client & server.

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/channel/SingleNioLoopProvider.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/channel/SingleNioLoopProvider.java
@@ -1,0 +1,26 @@
+package io.reactivex.netty.channel;
+
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
+
+/**
+ * An implementation of {@link RxEventLoopProvider} that returns the same {@link EventLoopGroup} instance for both
+ * client and server.
+ *
+ * @author Nitesh Kant
+ */
+public class SingleNioLoopProvider implements RxEventLoopProvider {
+
+    private final EventLoopGroup eventLoop = new NioEventLoopGroup(0/*means default in netty*/,
+                                                                   new RxDefaultThreadFactory("rx-selector"));
+
+    @Override
+    public EventLoopGroup globalClientEventLoop() {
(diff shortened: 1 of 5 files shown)


allenxwang:
The RxEventLoopProvider is static in RxNetty. So it means it can be easily set many times from application code. It is possible that in some code path, it is set but later gets overridden by a third party code and by the time the the client/server is created, it is not using the expected EventLoopGroup.

Should we restrict that it can be set only once?

NiteshKant:
This EventLoopProvider is a global provider which is used when the eventloop is not specified for a client or server instance. In general a global state should be set at the start of an application and not repeatedly during program execution. 
If I see the pros of putting this limitation that it can be set once, is just that unknowingly people will not be able to step on each other's feet. The con though is that it will enforce strict ordering of calls to update this provider. In an environment where an application constitutes of multiple libraries, strict ordering requirements like this become too restrictive.
I would like to consider this as something that hardly anyone would have to set and whoever sets it knows what he/she is doing.

rx-netty/src/main/java/io/reactivex/netty/channel/SingleNioLoopProvider.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * An implementation of {@link RxEventLoopProvider} that returns the same {@link EventLoopGroup} instance for both
+ * client and server.
+ *
+ * @author Nitesh Kant
+ */
+public class SingleNioLoopProvider implements RxEventLoopProvider {
+
+    private final EventLoopGroup eventLoop = new NioEventLoopGroup(0/*means default in netty*/,
```
benjchristensen, line 14: The default of 0 results in more threads than I'd expect. It was double the number of cores I think. 

Shouldn't it be the number of cores?

``` java
Runtime.getRuntime().availableProcessors()
```

NiteshKant, line 14: Yes, the default in netty is double the number of cores.

I wanted to stick to netty's defaults till the time we have benchmarked what is a good default for most applications. After we benchmark and have data, we can change this default.

Issue #82, opened 2014-03-25
AbstractQueueBasedChannelPool issues duplicate channel lifecycle events


NiteshKant:
The lifecycle of a channel in netty is defined as 

channelRegistered -> channelActive -> channelInactive -> channeUnregistered

However, going by the code here:

https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/client/pool/AbstractQueueBasedChannelPool.java#L150

The lifecycle will be:

channelRegistered -> channelActive -> channelRegistered -> channelActive -> channelInactive -> channeUnregistered

for the channel that gets reused once.

This will violate the contract of channel lifecycle and hence create confusion for people adding lifecycle handlers to the pipeline.

On a related note, why do we have to re-initialize the pipeline here? As discussed here: https://github.com/Netflix/RxNetty/issues/20 I thought we would not worry about resetting the pipeline and channel states on connection reuse?

allenxwang:
We need to re-initialize pipeline for the following reason:
- Make sure the new Observer is added at the end of the pipeline
- Make sure that custom objects kept in ChannelHandler will no longer be effective (since new instance of ChannelHandler is installed)
- Make the channel reusable for another RxClient that has different pipeline configuration, but talks to the same server.  This is the requirement from higher level client that utilizes the load balancer and share a connection pool for all servers in the farm.

Otherwise, we have to make sure all objects in the ChannelHandler is "resettable" and reset them when a channel is reused, which will make the code complicated.

allenxwang:
Also, since during re-initialization, all channel handlers are removed and new ones are installed, they will not get duplicated channel events. 

https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/client/pool/AbstractQueueBasedChannelPool.java#L146

Specifically, for channel handlers removed during re-initialization, they got:

channelRegistered -> channelActive (assuming the channel was not closed)

For new handlers installed, they will get the same.

I feel this is a simple way to decouple RxClient from handling channel from connection pool as it does not need to do anything special (other than setting an attribute) after getting the channel from the pool.

NiteshKant:
> > > Also, since during re-initialization, all channel handlers are removed and new ones are installed, they will not get duplicated channel events.

[...]

Issue #83, opened 2014-03-25
Synchronized methods in AbstractQueueBasedChannelPool


NiteshKant:
This could lead to hot locks specially if executed within a completely async application and gets executed in an eventloop.

For the methods in question, 

1) getTotalChannelsInPool() can just be a best effort and hence can remove synchronization. 
2) setMaxTotal() uses a semaphore underneath so does not need to be synchronized, rite?

allenxwang:
getTotalChannelsInPool() is only used for monitoring purpose which should happen infrequently in background thread. I cannot imagine why it would be used for a normal connection request. 

I also expect that set should be called very rarely, possibly several times a day, and in a background thread.

If get and set is not synchronized and set is in process, getTotalChannelsInPool() may return an incorrect number which could be a dramatic surge or drop that exceeds the expected range. This might cause false alarm or confusion.

allenxwang:
For set method, if two threads try to set the max at the same time and the method is not synchronized, maxTotal and maxConnectionsLimit may get out of sync. This will cause the get method return the wrong value for extended period of time until the next set.

NiteshKant:
> > > For set method, if two threads try to set the max at the same time and the method is not synchronized, maxTotal and maxConnectionsLimit may get out of sync. This will cause the get method return the wrong value for extended period of time until the next set.

Since, AdjustableSemaphore is defined by us, the instance variable AbstractQueueBasedChannelPool.maxTotal seems redundant to me, as you can provide a getter in AdjustableSemaphore to get this value. Positive is that you don't have to worry about consistency between these values.

> > > getTotalChannelsInPool() is only used for monitoring purpose which should happen infrequently in background thread. I cannot imagine why it would be used for a normal connection request.
> > > I also expect that set should be called very rarely, possibly several times a day, and in a background thread.

It is a public method, we should not make assumptions about its usage.

> > > If get and set is not synchronized and set is in process, getTotalChannelsInPool() may return an incorrect number which could be a dramatic surge or drop that exceeds the expected range.

I think the question here is what are the repercussions of an incorrect value in the scenario when these calls are interleaved. Yes, it may give an incorrect value without synchronization but how big is that con as opposed to having locks in a publicly exposed methods. If we have alarms ringing for one instance of an incorrect value (which is fixed in the next call to the method), we have a different problem to solve.

I leave the decision to you, to me it does not seem necessary.

NiteshKant:
After changes in pull request #95, we no more need these synchronized methods.

Issue #81, opened 2014-03-25
Connection pool shutdown


NiteshKant:
Currently, the ChannelPool implementation does not provide a way to shutdown the pool. After we resolve issue #80 of providing a shutdown for the client, we should also be able to shutdown the associated pool.
If we do not do this, we will end up with unused opened connections, if a client is shutdown and the process is still alive.

This should possibly also close the "in-use" connections and not just the idle ones.

NiteshKant:
Pull request #95 fixes this issue. However, it only closes idle connections and every released connection post the shutdown.
We may want to provide a shutdownNow() method sometime later that closes the checked-out connections too. This may be required for a force close for long running connections.

Issue #80, opened 2014-03-25
RxClient should provide a way to shutdown


NiteshKant:
This is specifically required since we have also implemented a connection pool and hence there should be a way to cleanup those connections when they are no longer required.

This should in turn shutdown the underlying pool.

PR #95, opened 2014-04-16
Connection pool refactoring

NiteshKant:
Made the following changes to ChannelPool:
- Improved the abstraction earlier provided by ChannelPool by converting it to a "ConnectionPool" to pool ObservableConnection. This fits in better with the rest of the code.
- Modified the interface to "assume" that a pool connects to one server. This also fits in better with the existing client abstraction.
- Provided a shutdown method on the pool. (Issue #80, #81)
- Modified the pool stats by not storing any monotonically increasing values as they overflow. Instead provided various callbacks which can be used to infer these stats.
- Removed the use of semaphores as they are easily replaced by an AtomicInteger. Also modified the method to update the max connections limit as atomic increment/decrements rather than a set, which is hard to make thread-safe and consistent without synchronization. (Discussed partially under issue #83)
- Fixed issue #82, #76 and #79

There are three pending tasks after this change:
- Provide a "cleanup thread" (using a threadpool) to cleanup idle connections.
- Provide a composite pool stats object encompassing multiple connection pools.
- Provide another PoolLimitDeterminationStrategy for max connections across pools which can be used to enforce connection limits over a pool of servers.


rx-netty/src/main/java/io/reactivex/netty/client/PooledConnection.java
```diff
@@ -0,0 +1,99 @@
...
+     * This makes sure that the underlying netty's channel is active as returned by
+     * {@link Channel#isActive()} and it has not passed the maximum idle time in the pool.
+     *
+     * @return {@code true} if the connection is usable.
+     */
+    public boolean isUsable() {
+        Boolean discardConn = getChannelHandlerContext().channel()
+                              .attr(ClientRequestResponseConverter.DISCARD_CONNECTION).get();
+
+        if (!getChannelHandlerContext().channel().isActive() || null != discardConn && discardConn) {
```
allenxwang: can be simplified as discardConn == Boolean.TRUE

rx-netty/src/main/java/io/reactivex/netty/channel/ObservableConnection.java
```diff
@@ -61,34 +61,40 @@ public boolean isCloseIssued() {
...
+        }
+    }
+
+    protected void cleanupConnection() {
+        cancelPendingWrites(true);
+        inputSubject.onCompleted();
+        ReadTimeoutPipelineConfigurator.removeTimeoutHandler(getChannelHandlerContext().pipeline());
+    }
+
+    protected Observable<Void> _closeChannel() {
```
allenxwang, line 78: why the "_" in the method name? Does it signal anything special?

rx-netty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -0,0 +1,262 @@
...
+
+        // Its legal to release after shutdown as it is not under anyones control when the connection returns. Its
+        // usually user initiated.
+
+        if (null == connection) {
+            return Observable.error(new IllegalArgumentException("Returned a null connection to the pool."));
+        }
+        try {
+            stateChangeListener.onReleaseAttempted();
+            if (isShutdown.get()) {
```
allenxwang: some logic is repeated in the if-elseif-else. Can be simplified as

```
if (!isShutdown.get() && connection.isUsable()) {
    ...
} else {
   ...
}
```

allenxwang:
I would really like to see some JUnit test for usage of CompositeStrategy since it will be heavily used by Ribbon and any client that utilizes a pool that can connect to different servers.

rx-netty/src/main/java/io/reactivex/netty/client/CompositeStrategy.java
```diff
@@ -0,0 +1,27 @@
+package io.reactivex.netty.client;
+
+/**
+ * @author Nitesh Kant
+ */
+public class CompositeStrategy extends CompositePoolStateChangeListener implements PoolLimitDeterminationStrategy {
```
allenxwang: Possibly rename to CompositePoolLimitDeterminationStrategy to make it more obvious.

rx-netty/src/main/java/io/reactivex/netty/channel/ObservableConnection.java
NiteshKant, line 78: Yeah I tend to add _ to non-public methods.

rx-netty/src/main/java/io/reactivex/netty/client/PooledConnection.java
NiteshKant: Sure, thanks!

rx-netty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
NiteshKant: Yes correct. will do.

rx-netty/src/main/java/io/reactivex/netty/client/CompositeStrategy.java
NiteshKant: Sure

NiteshKant:
@allenxwang I have added tests for PoolLimitDeterminationStrategy's various implementation. Thanks as there was actually a bug in there :)

rx-netty/src/main/java/io/reactivex/netty/client/MaxConnectionsBasedStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+        /**
+         * As opposed to limitEnforcer.incrementAndGet() we follow this model as this does not change the limitEnforcer
+         * value unless there are enough permits.
+         * If we were to use incrementAndGet(), in case of overflow (from max allowed limit) we would have to decrement
+         * the limitEnforcer. This may show temporary overflows in getMaxConnections() which may be disturbing for a
+         * user. However, even if we use incrementAndGet() the counter corrects itself over time.
+         * This is just a more semantically correct implementation with similar performance characterstics as
+         * incrementAndGet()
+         */
+        for (;;) {
```
allenxwang, line 43: I am wondering in high concurrency, some requests will end up having to spend quite some time in the loop. Or, theoretically, a thread can never exist the loop - just before limitEnforcer.compareAndSet(currentValue, newValue) for that request is going to be executed, the thread is suspended and another thread gets a chance to modify limitEnforcer.

NiteshKant, line 43: Yes but unfortunately, implementing atomic increments based on CAS are always like that. For reference if you look at jdk Semaphore implementation for tryAcquire() you will see similar code:

```
            for (;;) {
                int available = getState();
                int remaining = available - acquires;
                if (remaining < 0 ||
                    compareAndSetState(available, remaining))
                    return remaining;
            }
```

or AtomicInteger.incrementAndGet()

```
        for (;;) {
            int current = get();
            int next = current + 1;
            if (compareAndSet(current, next))
                return next;
        }
```

I am open to suggestions if you have a better way to do this.

NiteshKant:
@allenxwang are you comfortable with me merging this now?

rx-netty/src/main/java/io/reactivex/netty/client/MaxConnectionsBasedStrategy.java
allenxwang, line 43: Can we fail fast if we have spend too many cycles in the loop, like 1000?

NiteshKant, line 43: I think the correct terminology will be "fail slow" considering 1000 iterations :)

We surely can add that but that would mean that 1000 time concurrently at the same time two threads ended up in the same point. Specifically that this kind of loops are all over the Atomic package, adding this safety net seems unnecessary to me.

rx-netty/src/test/java/io/reactivex/netty/client/PoolLimitStrategyTest.java
```diff
@@ -0,0 +1,59 @@
...
+        Assert.assertFalse("Invalid permit acquire success.", strategy.acquireCreationPermit());
+
+        strategy.onNext(PoolInsightProvider.StateChangeEvent.OnConnectionEviction);
+
+        Assert.assertEquals("Unexpected available permits.", 1, strategy.getAvailablePermits());
+        Assert.assertTrue("Permit not available after release.", strategy.acquireCreationPermit());
+    }
+
+    @Test
+    public void testCompositeStrategy() throws Exception {
```
allenxwang, line 31: I am wondering if we should write another test case that test the normal situation that the global permits are greater than the local permits. Also some tests showing that max connections are resettable would be good.

allenxwang:
I feel it make sense to have a pool sub sub package under client since pool is not a required component for client.

allenxwang:
Do you have an example I can I subscribe to pool state change without the direct access to pool object itself? Is it through RxClientImpl.stateChangeObservable()? What if somebody creates the client using the builder and does not have access to the impl class?

rx-netty/build.gradle
```diff
@@ -38,6 +38,8 @@ dependencies {
     compile 'io.netty:netty-handler:4.0.14.Final'
     compile 'io.netty:netty-codec-http:4.0.14.Final'
     compile 'com.netflix.rxjava:rxjava-core:[0.17,)'
+    compile 'org.slf4j:slf4j-api:1.7.6'
+    compile 'com.netflix.numerus:numerus:1.+'
```
allenxwang, line 42: should be testCompile?

NiteshKant, line 42: Why? numerus is used for LongAdder in pool stats. slf4j is used for logging

NiteshKant:
> > > I feel it make sense to have a pool sub sub package under client since pool is not a required component for client.

How will having a different package name do anything with it being required or not?
The reason why it isn't a different package is that they share some package private classes/methods which are internal implementation detail.

NiteshKant:
> > > Do you have an example I can I subscribe to pool state change without the direct access to pool object itself? Is it through RxClientImpl.stateChangeObservable()? What if somebody creates the client using the builder and does not have access to the impl class?

RxClient interface provides this method (by virtue of extending it from PoolInsightProvider)

rx-netty/src/test/java/io/reactivex/netty/client/PoolLimitStrategyTest.java
NiteshKant, line 31: Sure, done.

rx-netty/src/main/java/io/reactivex/netty/client/PoolInsightProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author Nitesh Kant
+ */
+public interface PoolInsightProvider {
+
+    /**
+     * Returns the {@link Observable} that emits any changes to the state of the pool as {@link StateChangeEvent}
+     *
+     * @return An {@link Observable} emitting all state change events to the pool.
+     */
+    Observable<StateChangeEvent> stateChangeObservable();
```
allenxwang: Can we rename this poolStateChangeObservable and PoolStateChangeEvent? When this method is invoked on RxClient, it will be confusing what state change is about.

NiteshKant: Since StateChangeEvent is an inner class (enum) inside PoolInsightProvider, I thought it was indicative enough with a shorter name.
Anyways, I have renamed it.

Issue #90, opened 2014-04-04
Server onError Handling


benjchristensen:
The server does not have default error handling when `Observable<Void>` calls onError.

Right now I have to manually do it or errors are swallowed:

``` java
                    return testEndpoint(request, response).onErrorFlatMap(error -> {
                        return writeError(request, response, "Unknown error: " + error.getMessage());
                    });
```

NiteshKant:
I guess if the method HttpServer.withErrorResponseGenerator() was public (which was the original intent :) )

https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpServer.java#L43

it would suffice, rite? 

One can define a global catch all error response generator using the above.

benjchristensen:
Yes I should be able to override it, but there should be a default like servlet containers that returns a status 500. Right now it just silently fails and swallows errors.

NiteshKant:
Actually, there is a default error generator that sets the status to 500. If you create a server like:

```
RxNetty.createHttpServer(9999, new RequestHandler<ByteBuf, ByteBuf>() {
                                              @Override
                                              public Observable<Void> handle(
                                                      HttpServerRequest<ByteBuf> request,
                                                      HttpServerResponse<ByteBuf> response) {
                                                  return Observable.error(new UnsupportedOperationException());
                                              }
                                          }).startAndWait();
```

You will get an error response as:

```
curl -v "http://localhost:9999"
* About to connect() to localhost port 9999 (#0)
*   Trying ::1...
* connected
* Connected to localhost (::1) port 9999 (#0)
> GET / HTTP/1.1
> User-Agent: curl/7.24.0 (x86_64-apple-darwin12.0) libcurl/7.24.0 OpenSSL/0.9.8y zlib/1.2.5
> Host: localhost:9999
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error
< Transfer-Encoding: chunked
< 
* Connection #0 to host localhost left intact
* Closing connection #0
```

However, there is a caveat. The error response generator is called if and only if, a response (partial or full) is not already written on that channel. So, if you instead use the following request handler:

```
                                          new RequestHandler<ByteBuf, ByteBuf>() {
                                              @Override
                                              public Observable<Void> handle(
                                                      HttpServerRequest<ByteBuf> request,
                                                      HttpServerResponse<ByteBuf> response) {
                                                  response.writeString("hello");
[...]

Issue #100, opened 2014-04-21
Race condition of pool idle connection cleanup task and acquiring connection


allenxwang:
In the code of IdleConnectionsCleanupTask:

```
    PooledConnection<I, O> idleConnection = iterator.next(); // 1
    if (!idleConnection.isUsable()) { // 2
           iterator.remove(); 
           discardConnection(idleConnection); 
    }
```

Follow the sequence below:
1. clean up thread gets an idle connection from `iterator.next()` at mark 1
2. clean up thread gets suspended
3. another thread acquires the same connection via `idleConnections.poll()`
4. clean up thread resumes, and at mark 2 `idleConnection.isUsable()` may return false since some time has elapsed 
5. clean up thread closes the connection, which is already in use in another thread

Possibly make sense to use an AtomicBoolean to indicate that a PooledConnection has been owned by a thread?

NiteshKant:
Good point.

Comparing it with an otherwise normally occurring scenario: 

Connection is closed by the OS when it is in use (this will primarily be due to server closing the connection).

Will the system behave any differently in these two scenarios? I believe no.

Now, can we avoid this scenario?

Let us take your possible fix. If we do have a flag "isUsed" in the connection, then the above code will convert to:

```
PooledConnection<I, O> idleConnection = iterator.next(); // 1
    if (!idleConnection.isUsable() && !idleConnection.isInUse()) { // 2
           iterator.remove(); 
           discardConnection(idleConnection); 
    }
```

(The isInUse() check can be inside isUsable() but this is just for discussion purpose)

Is the above fix any different than what is there currently? I think no because the threads can be interleaved so that the connection acquire is between isInUse() check and iterator.remove()/discard call. So it is essentially the same situation.

I think the only way we can ensure that it is not used while in the iterator loop is that instead of having an iterator, we do a drain() from the queue and then add back the connections which are usable. Considering normal usage pattern of highly used pools, I guess if this concurrency situation is to be considered, we should assume that the pool is in frequent use and hence most of the connections are _not_ idle. Hence, we should optimize for connections that are not idle for long (i.e. not expired) than the situation in which they are. The completely safe situation optimizes for the situation when most of the connections are about to be expired. The current one hopes that there are less such occurrences.

allenxwang:
Would this fix the problem, assuming `owned` is an AtomicBoolean:

For cleaning thread:

``` java
PooledConnection<I, O> idleConnection = iterator.next(); // 1
[...]

PR #103, opened 2014-04-22
Fixes issue #90 #100

NiteshKant:
Issue #90: Added a basic HTML page dumping the error stack for the "catch-all" error handler.
Issue #100: Added a package-private claim() method on PooledConnection that provides a way to claim a connection for use, which is returned on close().

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/client/PooledConnection.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/client/PooledConnection.java
@@ -7,6 +7,8 @@ import io.reactivex.netty.protocol.http.client.ClientRequestResponseConverter;
 import rx.Observable;
 import rx.subjects.PublishSubject;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 /**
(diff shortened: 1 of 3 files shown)


rx-netty/src/main/java/io/reactivex/netty/client/PooledConnection.java
```diff
@@ -33,10 +38,12 @@ public PooledConnection(ChannelHandlerContext ctx, ConnectionPool<I, O> pool, lo
 
     @Override
     public Observable<Void> close() {
+        acquiredOrSoonToBeDiscarded.compareAndSet(true, false); // There isn't anything else to be done here.
```
allenxwang, line 41: could it be just `acquiredOrSoonToBeDiscarded.set(false)`?

NiteshKant, line 41: I think the semantically correct code is acquiredOrSoonToBeDiscarded.compareAndSet(true, false);

PR #108, opened 2014-04-29
Udp support

NiteshKant:
Added UDP support for RxNetty.

The UDP client follows the same interface as other clients. This is a little counter-intuitive (as UDP is connectionless and there is a method connect() ) but provides a handy way to write/read UDP packets to/from a particular server as opposed to having to be done manually for each request.

Added a Java client and server example to understand the usage.

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/channel/DefaultChannelWriter.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/channel/DefaultChannelWriter.java
@@ -2,6 +2,7 @@ package io.reactivex.netty.channel;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
 import io.reactivex.netty.protocol.http.MultipleFutureListener;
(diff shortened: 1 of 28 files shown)


rx-netty/src/main/java/io/reactivex/netty/channel/DefaultChannelWriter.java
```diff
@@ -93,13 +94,17 @@ public ByteBufAllocator getAllocator() {
         return ctx.alloc();
     }
 
+    public ChannelHandlerContext getChannelHandlerContext() {
+        return ctx;
+    }
+
     protected ChannelFuture writeOnChannel(Object msg) {
-        ChannelFuture writeFuture = ctx.channel().write(msg); // Calling write on context will be wrong as the context will be of a component not necessarily, the head of the pipeline.
+        ChannelFuture writeFuture = getChannel().write(msg); // Calling write on context will be wrong as the context will be of a component not necessarily, the head of the pipeline.
```
normanmaurer, line 102: nitpicking... the  "....the tail of the pipeline". Writes always start at the tail (in fact all outbound operations).

rx-netty/src/main/java/io/reactivex/netty/protocol/udp/client/UdpClient.java
```diff
@@ -0,0 +1,36 @@
+package io.reactivex.netty.protocol.udp.client;
```
normanmaurer, line 1: Missing license headers ;) Also in other classes

NiteshKant, line 1: Thanks! I created a different PR #109 for updating the copyright.

rx-netty/src/main/java/io/reactivex/netty/channel/DefaultChannelWriter.java
NiteshKant, line 102: Aah, With head I was actually referring to the "start" of the pipeline for outbound operations, which actually is the tail! English is a funny language :)
I will change the comment anyways, thanks!

Issue #117, opened 2014-05-10
RxClientImpl mutates the bootstrap per connection


NiteshKant:
RxClientImpl via [ClientChannelFactoryImpl](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/client/ClientChannelFactoryImpl.java#L57) mutates the bootstrap object, updating the ChannelInitializer. This is wrong as the bootstrap is used across multiple connections and hence different initializers steps on each others steps.
This eventually ends up using the same [ClientConnectionHandler](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/client/ClientConnectionHandler.java) across multiple connections and thus one subscriber of [RxClient.submit()](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/client/RxClient.java) receives multiple onNext() calls.
In HTTP client this results in the same request being written multiple times on different connections.

Issue #114, opened 2014-05-06
Remove numerus dependency from core


NiteshKant:
Currently numerus is being used for LongAdder an alternate implementation of which is available in netty-common which we already depend on.

NiteshKant:
We were using LongAdder from numerus. Now using netty's internal backport of the same.

headinthebox:
Cool; that caused me some heartburn before.

NiteshKant:
:) yeah it wasn't required. Keep the dependency tree lean is the mantra!

Issue #113, opened 2014-05-06
Implement Redirect as an Rx Operator?


NiteshKant:
Currently the follow redirect logic is implemented in a [child class](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/FollowRedirectHttpClientImpl.java) of HttpClientImpl.

One basic issue with the redirect implementation is that it assumes that the client configuration for client of the redirect location is the same as the client which got the first request. 

If we implement this as an operator, it can also be used externally for clients which wants to use a different client for issuing redirect requests.

Above is a concrete advantage of this change apart from the fact that a child class just for redirect seems unnecessary.
By implementing this as an operator, we can chain redirect functionality on the main observable returned by HttpClientImpl.submit()

NiteshKant:
Implemented redirect as [RedirectOperator](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/RedirectOperator.java). HttpClientImpl uses this operator if followRedirect is not opted out.
If someone requires a different redirect implementation, the behavior is best altered by implementing a [RedirectHandler](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/RedirectOperator.java#L91). In such a case, one should disable followRedirect [HttpClientConfig.setFollowRedirect(false)](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/HttpClient.java#L70) and use the RedirectOperator with their own RedirectHandler in a lift on the Observable returned by HttpClient.submit()

Example:

```
HttpClientRequest<ByteBuf> request = HttpClientRequest.createGet("http://localhost:8080/hello");

        RxNetty.createHttpRequest(request,
                                  new HttpClient.HttpClientConfig.Builder().setFollowRedirect(false).build())
               .lift(new RedirectOperator<ByteBuf, ByteBuf>(request, new RedirectOperator.RedirectHandler<ByteBuf, ByteBuf>() {

                   @Override
                   public Observable<HttpClientResponse<ByteBuf>> doRedirect(
                           RedirectionContext context, HttpClientRequest<ByteBuf> originalRequest,
                           HttpClientResponse<ByteBuf> redirectedResponse) {
                       // TODO: implement redirect
                   }

                   @Override
[...]

Issue #88, opened 2014-04-02
Move host and port to RxClient method signature


allenxwang:
Currently host and port are required arguments for the constructor and instance variables for RxClientImpl. I am proposing moving them to method signature like connect() or submit(). This will have the following benefit:
- Makes RxClient reusable when it needs to connect to a different server but uses the same client and pipeline configuration
- Makes it possible to have APIs that support full HTTP URI that includes host and port
- Provides an easier way to integrate with Ribbon load balancer which provides host and port.

benjchristensen:
Right now it is this awkward design:

``` java
HttpClient<ByteBuf, ByteBuf> client = RxNetty.createHttpClient("localhost", 8080);

client.submit(HttpClientRequest.createGet("/"));
```

I assume you mean we'd instead do something like:

``` java
HttpClient<ByteBuf, ByteBuf> client = RxNetty.createHttpClient();

client.submit(HttpClientRequest.createGet("http://localhost:8080/")));
// which can be aliased for the GET case as
client.get("http://localhost:8080/"));
```

NiteshKant:
The model that RxClient has now is that an RxClient implementation must connect to either one (the current implementation) or a group of "homogeneous" servers. 

I guess there are two distinct type of users of RxClient (with respect to this discussion):
- _Casual users_: Users who are playing with RxNetty with an intent of experimentation/learning.
-  _Production applications_: Applications based on higher level clients like ribbon. 

In ribbon as we do today, we create ribbon clients for a group of homogeneous servers. Grouping can be based on VIPs, applications, clusters, asgs, etc. but all the servers are homogeneous. What this essentially means that while executing a request, the user does not specify a host/port. All they do is; ask ribbon to execute this request on "any server" in the farm. This kind of groupings inside clients is certainly natural because the client configurations (eg: connection timeouts, read timeouts, etc.) are best configured for a homogeneous set of servers.
RxClient follows this model targeting mainly users in the second category above.

Now, ribbon (atleast inside Netflix) marries these two usecases by giving the ability to pass a VIP address or a host in the URL, but that comes with a complexity of analysing the URL, trying to match with a VIP address or other criterion to decide if this is a host or not.

I have tried to stay away from this approach sighting simplicity in design. Having said that, if we were to create a client that can connect to a set of heterogenous servers, we can surely do that by creating an abstraction on top of these clients with a slightly different API (of passing host/port) which internally is a composite of individual RxClients specific to different host/port combinations.

[...]

PR #120, opened 2014-05-17
Fix for issue #117 #114 #113 #88

NiteshKant:
Issue #117: This was a major issue with the RxClient implementation which was mutating (updating ChannelInitializer) in the bootstrap for every connect. The reason for doing this was to have a ClientLifecycleListener provide connection events. This design was naively taken from the RxServer but does not apply for the client as the LifecycleListener requires an observer which changes for the client per connection.
This change fixes this issue by removing the newConnection notification responsibility from the LifecycleListener and have it done by the ClientChannelFactory. This also enabled to remove a few other redundant constructs like the AbstractChannelFactory which was unnecessarily complex. So, the code looks much cleaner now :)

Issue #114: rx-netty core was using numerus for LongAdder implementation. Now it uses netty's internal backport of LongAdder.

Issue #113: Rewrote the Redirect implementation as an Rx operator. This makes the code much cleaner and provides a very useful feature of being able to elect which HttpClient implementation must be used for executing redirects. This was a major limitation of the previous approach.

Issue #88: Provide an implementation of CompositeHttpClient which is a composite of HttpClients per ServerInfo. This provides additional submit() methods which also takes a host, port. By default this client submits requests to localhost.
RxNetty now has two flavors of createHttpRequest() that takes a HttpClientRequest with the full URI set and uses an appropriate HttpClient instance to execute this request.
Now, one can use a shorthand:

```
RxNetty.createHttpRequest(HttpClientRequest.createGet("http://localhost:8080/hello"))
```

to submit a request with full URI which was not previously possible.


rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/DefaultRedirectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    protected Observable<HttpClientResponse<O>> redirect(HttpClientRequest<I> redirectRequest) {
+        return client.submit(redirectRequest, config);
+    }
+
+    @Override
+    public boolean requiresRedirect(RedirectionContext context, HttpClientResponse<O> response) {
+        int statusCode = response.getStatus().code();
+        return statusCode >= 301 && statusCode <= 308;
```
allenxwang: 304 is for HTTP caching, not for redirect.

benjchristensen:
How about for the most common cases we have overloads like this?

``` java
RxNetty.createHttpGet("http://localhost:8080/hello")
```

rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/DefaultRedirectHandler.java
NiteshKant: Thanks! I updated the code to only redirect for 301,302,303, 307, 308.

NiteshKant:
@benjchristensen yes I will add that method. The caveat being that if the user wants to add any header, the HttpClientRequest builder has to be used.
I will also add similar method for DELETE.
Shorthands for PUT & POST becomes a bit more complex as we have a few ways of specifying the content. So, I will stay away from them as of now.

I personally feel we should give a bit more thought on making this more usable by providing a more fluent DSL more on the lines you suggested offline i.e. by having an extension to Observable. I will put together my thoughts on that and open a different issue.

Issue #126, opened 2014-05-23
HttpServer not compliant to HTTP 1.0 requests


NiteshKant:
There are multiple issues with the current HttpServer implementation.
- According to [RFC 2145](http://tools.ietf.org/html/rfc2145#section-2.3) a server should return the maximum compatible HTTP minor version (which in our case will be 1.1) for any request 1.0 or 1.1. Instead currently, the server sends a 1.0 response for 1.0 requests.
- The server does not send a "Connection" header with value as "keep-alive" for client requested Keep-Alive and "Close" with no keep-alive requests. ([HTTP spec reference](http://www.w3.org/Protocols/HTTP/1.1/draft-ietf-http-v11-spec-01.html#Connection))
- Even if the transfer encoding is not chunked, the server sends a last http content.
- The transfer encoding chunked is set only for keep-alive connections even though the response is always sent as chunked.

PR #128, opened 2014-05-26
Fixes issue #126

NiteshKant:
HttpServer now handles HTTP 1.0 requests as per the spec.


NiteshKant:
> > > RxNetty-pull-requests #53 FAILURE
> > > Looks like there's a problem with this pull request

Strange that the build failed due to maven not reachable.

```
* What went wrong:
Could not resolve all dependencies for configuration ':rx-netty:testCompile'.
> Could not resolve junit:junit:4.10.
  Required by:
      com.netflix.rxnetty:rx-netty:0.3.6
   > Could not GET 'http://repo1.maven.org/maven2/junit/junit/4.10/junit-4.10.pom'.
      > No route to host
```

PR #133, opened 2014-06-02
Minor client changes.

NiteshKant:
1) Modified RxClientImpl.connect() to do a take(1) on the underlying source so that in no case it will emit two onNext().
2) Modified RequestProcessingOperator to use SubscriptionList instead of CompositeSubscription. Instead of creating a new subscription on every onNext(), it is done once and used by the subscriber.
3) Provided an override to remove all connection pooling for a client.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
 # RxNetty Releases #
 
+
+### Version 0.3.6 ###
+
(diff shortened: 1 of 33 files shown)


NiteshKant:
I am looking at why all pull request builds are getting aborted.

NiteshKant:
There were some infinite waits for server shutdown in some cases which were causing the build in cloudbee to hang, strangely so only for pull request builds. I have replaced these waits with timed waits.

Issue #106, opened 2014-04-27
HTTPS Support


mathieucarbou:
Hello,
Does RxNetty Http client can connect to https websites ? I do not see any doc or sample relating to https.
Thank you!

NiteshKant:
Netty supports SSL via [SSLHandler](https://github.com/netty/netty/blob/master/handler/src/main/java/io/netty/handler/ssl/SslHandler.java) so RxNetty also does support SSL.

You can create a PipelineConfigurator implementation that adds an SSLHandler in the pipeline much as [this example](https://github.com/netty/netty/tree/master/example/src/main/java/io/netty/example/securechat) in netty demonstrates. In order to understand RxNetty's PipelineConfigurators have a look at the [pipeline package](https://github.com/Netflix/RxNetty/tree/master/rx-netty/src/main/java/io/reactivex/netty/pipeline)

Having said the above, I think we should add an HTTPs example. I will add the same.

tbak:
To add SSL/TLS support to RxNetty, the RxServer/RxClient pipeline setup must be made more flexible.

SSL handling is supported by Netty's SslHandler ChannelHandler. It enforces a three step process for full communication setup:
1. Start with basic pipeline configuration and SSL ChannelInitializer in the pipeline
2. On channel registration inject SslHandler as a first handler
3. On SslHandshakeCompletionEvent event activate RxNetty's ConnectionHandler (to ensure nothing is written to the channel prior to handshake completion)

The first two steps can be done easily now. What is less obvious is how to prevent ObservableConnection from firing write events into the channel prior to getting handshake confirmation event.

Having said that here is a working prototype for TCP protocol without waiting for handshake complete event: https://github.com/tbak/RxNetty/commit/4c55b2122e1816cf464eaca079ea5b532584844e

Any read or write to the channel will enforce handshake negotiation in a first place. I am not sure if we can rely on it, or strictly follow the rules written in SslHandler documentation.

tbak:
First version with proper handling of the handshake event is available here: https://github.com/tbak/RxNetty/commits/features/ssl

tbak:
SSL support is now implemented. Most of the changes are in this commit: https://github.com/Netflix/RxNetty/commit/07c39de2a9e3c8b490c1b68dd5c2462ebaf3e2ae.

PR #147, opened 2014-06-19
Add SSL support (issue #106).

tbak:


rx-netty/src/main/java/io/reactivex/netty/RxNetty.java
```diff
@@ -97,6 +98,14 @@ private RxNetty() {
         return newTcpServerBuilder(port, connectionHandler).pipelineConfigurator(pipelineConfigurator).build();
     }
 
+    public static <I, O> RxServer<I, O> createSslInsecureTcpServer(final int port, PipelineConfigurator<I, O> pipelineConfigurator,
+                                                        ConnectionHandler<I, O> connectionHandler) {
+        return newTcpServerBuilder(port, connectionHandler)
+                .pipelineConfigurator((PipelineConfigurator<I, O>) PipelineConfigurators.sslInsecureServerConfigurator())
```
NiteshKant: Instead of casting, you can use PipelineConfigurators.&lt;I, O&gt;sslInsecureServerConfigurator()

rx-netty/src/main/java/io/reactivex/netty/pipeline/ssl/SslConfiguration.java
```diff
@@ -0,0 +1,288 @@
...
+public class SslConfiguration {
+
+    private boolean clientMode;
+    private boolean wantClientAuth;
+    private String[] protocols;
+    private String[] cipherSuite;
+    private final KeyManagerFactory keyManagerFactory;
+    private final TrustManagerFactory trustManagerFactory;
+
+    SslConfiguration(boolean userMode, boolean wantClientAuth, String[] protocols, String[] cipherSuite,
```
NiteshKant: I think we are doing more than required here.
In order to provide this SslConfiguration as an abstraction over the SSLEngine, we have to replicate code from netty; `PemReader` and some of `JdkSslContext` which does not seem correct.

IMO, we should simply have someone pass `SSLEngine` to create a `SslPipelineConfigurator`.
One can use netty's `SSLContext` to create an instance of `SSLEngine` and pass it to us. If the `SSLContext` class is not "simple" for our needs, we should create facilities to simplify that but the core API (RxNetty & PipelineConfigurators) should just take an `SSLEngine`

rx-netty/src/main/java/io/reactivex/netty/channel/ObservableConnection.java
```diff
@@ -36,7 +36,8 @@
     public ObservableConnection(final ChannelHandlerContext ctx) {
         super(ctx);
         inputSubject = PublishSubject.create();
-        ctx.fireUserEventTriggered(new NewRxConnectionEvent(inputSubject));
+        ChannelHandlerContext firstContext = ctx.pipeline().firstContext();
```
NiteshKant, line 39: I did not understand this code.
In the `ClientChannelFactoryImpl` we pass `pipeline.lastContext()` and over here take `ctx.pipeline().firstContext()`. So, what is this change achieving? :)

rx-netty/src/main/java/io/reactivex/netty/client/ClientChannelFactoryImpl.java
```diff
@@ -65,15 +61,27 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
+                    final ObservableConnection<I, O> newConnection = connectionFactory.newConnection(ctx);
                     ChannelHandler lifecycleHandler = pipeline.get(RxRequiredConfigurator.CONN_LIFECYCLE_HANDLER_NAME);
-                    if (null != lifecycleHandler) {
+                    if (null == lifecycleHandler) {
+                        onNewConnection(newConnection, subscriber);
+                    } else {
                         @SuppressWarnings("unchecked")
                         ConnectionLifecycleHandler<I, O> handler = (ConnectionLifecycleHandler<I, O>) lifecycleHandler;
-                        handler.setConnection(newConnection);
+                        ChannelHandler sslHandler = pipeline.get(SslHandler.class);
```
NiteshKant: I think this code can be simplified by getting `SSLHandler.handshakeFuture()` and adding a listener to it, instead of registering an action with the `ConnectionLifecycleHandler`

rx-netty/src/main/java/io/reactivex/netty/client/ConnectionLifecycleHandler.java
```diff
@@ -38,4 +41,16 @@ public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
             connection = newConnection;
         }
     }
+
+    void setConnectionAction(Action0 connectionSetupAction) {
```
NiteshKant: If we use `SSLHandler.handshakeFuture().addListener()` in `ClientChannelFactoryImpl` this code will not be required.

rx-netty/src/main/java/io/reactivex/netty/pipeline/PipelineConfigurators.java
```diff
@@ -88,6 +90,21 @@ private PipelineConfigurators() {
     }
 
     /**
+     * This configurator creates self signed certificate for the server. Its primarily for testing purposes.
+     */
+    public static <I, O> PipelineConfigurator<I, O> sslInsecureServerConfigurator() {
```
NiteshKant: Based on the previous comment of using `SSLEngine` this should just give a way to create a `PipelineConfigurator` given an `SSLEngine`

gradle.properties
```diff
@@ -1,4 +1,4 @@
 version=0.3.6
-netty_version=4.0.14.Final
+netty_version=4.0.20.Final
```
NiteshKant, line 2: We would have to add this release to the Netflix artifactory before we can merge this.

rx-netty/src/main/java/io/reactivex/netty/channel/ObservableConnection.java
tbak, line 39: We pass lastContext to ConnectionObservable, so all writes will go through whole pipeline.
fireUserEventTriggered belongs to the ChannelInboundHandler interface, so we have to start from the head of the pipeline. With the previous code, we pointed to ObservableAdapter context (the last one in pipeline), and fireUserEventTriggered which was sent to next handler in chain (tail) had no effects.

NiteshKant, line 39: Go it! Sounds reasonable.

allenxwang:
Currently Ribbon's SSL support involves with creating SSLSocketFactory from configurations. How can I create the a PipelineConfigurator using a SSLSocketFactory?

graemerocher:
Sorry I'm wondering where did this pull request go? It was merged into master but there is no master branch. Am I missing something obvious?

NiteshKant:
@graemerocher `master` was renamed to `0.4.x` to better indicate the version. SSL is supported both in 0.4.x and 0.5.x artifacts.

PR #159, opened 2014-07-01
Fix for issue with HTTP pool connection and request content observable.

tbak:


NiteshKant:
@tbak I have done the change for issue #160 in a different PR #161

Issue #164, opened 2014-07-02
Server HttpConnectionHandler optimization (remove flatmap)


NiteshKant:
Based on our benchmarking the use of `Observable.flatMap()` inside `HttpConnectionHandler` [here](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpConnectionHandler.java#L60) incurs an unnecessary overhead.
The below flamegraph shows the overhead:

![httpconnhandlerflatmap](https://cloud.githubusercontent.com/assets/820005/3452670/0af22336-01b8-11e4-996d-13bce1fa62c4.png)

In this particular instance, `flatMap()` does not seem to be useful, it can easily be replaced by a custom operator which connects the `connection.getInput()`'s onComplete() and onError() to the subscriber of `HttpConnectionHandler.handle()` The result of `RequestHandler.handle()` does not impact the subscriber of `HttpConnectionHandler.handle()`

NiteshKant:
Here is the flamegraph after this fix:

![httpconnhandlerflatmapremoved](https://cloud.githubusercontent.com/assets/820005/3466726/bc40d02e-0285-11e4-80fd-cc1b28981d53.png)

PR #165, opened 2014-07-02
Fixes issue #164

NiteshKant:
Removed the use of flatMap() with a custom operator.


NiteshKant:
I am punting this PR for now. Will re-issue if/when we see the performance gain.

Issue #154, opened 2014-06-30
HttpServerRequest.getPath returns complete URL


tbak:
Calling getPath method in HTTP hello world example returns "http://localhost:8090/hello", while it should be only "/path".

NiteshKant:
@tbak the problem is not with the server in the helloworld example. It is with the client.
If you curl the endpoint, you will see the following output on the server:

```
New request received
GET /hello HTTP/1.1
User-Agent: curl/7.24.0 (x86_64-apple-darwin12.0) libcurl/7.24.0 OpenSSL/0.9.8y zlib/1.2.5
Host: localhost:8090
Accept: */*
Content-Length: 0
```

whereas if you use the `HelloWorldClient` you will see the following output on the server:

```
New request received
GET http://localhost:8090/hello HTTP/1.1
Host: localhost
User-Agent: RxNetty Client
Content-Length: 0
```

Now, if you enable the wire logging for the server, the `HelloWorldClient` request shows like this:

```
ERROR [rx-selector-1-2] (Slf4JLogger.java:161) - [id: 0x6a2e2275, /127.0.0.1:61710 => /127.0.0.1:8090] RECEIVED(89B)
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 47 45 54 20 68 74 74 70 3a 2f 2f 6c 6f 63 61 6c |GET http://local|
|00000010| 68 6f 73 74 3a 38 30 39 30 2f 68 65 6c 6c 6f 20 |host:8090/hello |
|00000020| 48 54 54 50 2f 31 2e 31 0d 0a 48 6f 73 74 3a 20 |HTTP/1.1..Host: |
|00000030| 6c 6f 63 61 6c 68 6f 73 74 0d 0a 55 73 65 72 2d |localhost..User-|
|00000040| 41 67 65 6e 74 3a 20 52 78 4e 65 74 74 79 20 43 |Agent: RxNetty C|
|00000050| 6c 69 65 6e 74 0d 0a 0d 0a                      |lient....       |
+--------+-------------------------------------------------+----------------+
```

As you can see above the request generated by the client is wrong.

tbak:
You are right. We have small inconsistency in RxNetty factory methods. They expect absolute URIs, so they can extract from them ServerInfo, but this URI is also passed in HTTP request in absolute form.

Crystark:
[...]

PR #171, opened 2014-07-03
Fix issue #154 - HttpClientRequest should always be passed relative URI.

tbak:
HttpClientRequest and its underlying Netty's HttpRequest assume that relative URI is provided as an argument.
To always fullfil this contract I propose to deprecate methods:
`RxNetty.createHttpRequest(HttpClientRequest<ByteBuf> request)`
`RxNetty.createHttpRequest(HttpClientRequest<ByteBuf> request, HttpClient.HttpClientConfig config))`
which require creation of HttpClientRequest with absolute URI (so server/port can be extrated from it),
and replace them with equivalent methods that accept server/port explicitely as method parameters.

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/RxNetty.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/RxNetty.java
@@ -25,6 +25,7 @@ import io.reactivex.netty.channel.RxEventLoopProvider;
 import io.reactivex.netty.channel.SingleNioLoopProvider;
 import io.reactivex.netty.client.ClientBuilder;
 import io.reactivex.netty.client.RxClient;
+import io.reactivex.netty.client.RxClient.ServerInfo;
 import io.reactivex.netty.metrics.MetricEventsListenerFactory;
 import io.reactivex.netty.pipeline.PipelineConfigurator;
 import io.reactivex.netty.protocol.http.client.CompositeHttpClient;
@@ -177,6 +178,19 @@ public final class RxNetty {
         return RxNetty.<I, O>newHttpClientBuilder(host, port).pipelineConfigurator(configurator).build();
     }
 
+    public static Observable<HttpClientResponse<ByteBuf>> createHttpRequest(String host, int port, HttpClientRequest<ByteBuf> request) {
+        RxClient.ServerInfo serverInfo = new ServerInfo(host, port);
+        return globalClient.submit(serverInfo, request);
+    }
+
+    public static Observable<HttpClientResponse<ByteBuf>> createHttpRequest(String host, int port,
+                                                                            HttpClientRequest<ByteBuf> request,
+                                                                            HttpClient.HttpClientConfig config) {
+        RxClient.ServerInfo serverInfo = new ServerInfo(host, port);
+        return globalClient.submit(serverInfo, request, config);
+    }
+
+    @Deprecated
     public static Observable<HttpClientResponse<ByteBuf>> createHttpRequest(HttpClientRequest<ByteBuf> request) {
         RxClient.ServerInfo serverInfo;
         try {
@@ -187,6 +201,7 @@ public final class RxNetty {
         return globalClient.submit(serverInfo, request);
     }
 
+    @Deprecated
     public static Observable<HttpClientResponse<ByteBuf>> createHttpRequest(HttpClientRequest<ByteBuf> request,
                                                                              HttpClient.HttpClientConfig config) {
         RxClient.ServerInfo serverInfo;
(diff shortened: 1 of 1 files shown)


NiteshKant:
@tbak the purpose of `RxNetty.createHttpRequest(HttpClientRequest<ByteBuf> request)`
`RxNetty.createHttpRequest(HttpClientRequest<ByteBuf> request, HttpClient.HttpClientConfig config))` was to provide an "easier" way to submit an HTTP request from a full URI.
These methods extract the server host and port from the request, defaulting the host/port if they do not exist in the URI.

The new methods you are suggesting has roughly the same creation cumbersomeness as:

``` java
RxNetty.createHttpClient("127.0.0.1", 8080).submit(HttpClientRequest.createGet("/hello"));
```

as opposed to:

``` java
RxNetty.createHttpRequest(HttpClientRequest.createGet("http://127.0.0.1:8080/hello"));
```

The former is away from the "normal" usage pattern that people have of creating a request from the full URI. This is the usecase, the existing methods are trying to solve.

tbak:
I understand that. There are still methods like createHttpGet(String uri),etc that provide the simple usage pattern.
Going for call like this:
RxNetty.createHttpRequest(HttpClientRequest.createGet("http://127.0.0.1:8080/hello"));
has no value over the
RxNetty.createHttpGet("http://127.0.0.1:8080/hello"));
which is more compact. You may need to use RxNetty.createHttpRequest to setup more HTTP request properties (set headers, cookies, content factory, etc), and that is already more complex scenario.
I think the bigger value here is with fulfilling the HttpClientRequest contract (create object with relative URIs only), and designing RxNetty factory methods around that.

NiteshKant:
> > > RxNetty.createHttpRequest(HttpClientRequest.createGet("http://127.0.0.1:8080/hello"));
> > > has no value over the
> > > RxNetty.createHttpGet("http://127.0.0.1:8080/hello"));

Indeed it does not but if we have this method, this can be used to make complex request started from the full URI. OTOH, if we just have `RxNetty.createHttpGet("http://127.0.0.1:8080/hello")` then the complex requests building is not additive to the simple full URI request submission.

To elaborate this, lets take an example of adding a header and a cookie. If we do not have the method `RxNetty.createHttpRequest(HttpClientRequest request)` which honors a full URI, one would have to do:

``` java
        RxNetty.createHttpClient("127.0.0.1", 8080)
               .submit(HttpClientRequest.createGet("/hello")
                                        .withHeader("Header1", "Val1")
                                        .withCookie(new DefaultCookie("mycookie", "cookieval")));
```

as opposed to:

``` java
        RxNetty.createHttpRequest(HttpClientRequest.createGet("http://127.0.0.1:8080")
                                                   .withHeader("Header1", "Val1")
                                                   .withCookie(new DefaultCookie("mycookie", "cookieval")));
```

The later is a more compact way of submitting requests and does not need people to change the approach if they have to add a header vis-a-vis just URI.

> > > I think the bigger value here is with fulfilling the HttpClientRequest contract (create object with relative URIs only), and designing RxNetty factory methods around that.

True but we can achieve this same goal by just updating the request URI in the passed `HttpClientRequest` instance, correct?

tbak:
> True but we can achieve this same goal by just updating the request URI in the passed HttpClientRequest instance, correct?

This is what I wanted to avoid, since in some contexts we will be expecting absolute and in some relative URIs when we create HttpClientRequest. For example:

``` java
RxNetty.createHttpClient("127.0.0.1", 8080)
    .submit(HttpClientRequest.createGet("http://netflix.com/hello"));
```

Which has higher priority? We could of course report this as an error.

NiteshKant:
It is a valid question as to what should a user expect to put as a URI in the `HttpClientRequest`

Let us outline the scenarios:
- `HttpClientRequest` has full URI and it is submitted directly to an `HTTPClient`.
- `HttpClientRequest` has full URI and it is submitted using `RxNetty.createHttpRequest` method flavors.

**Now, the contract for any HttpClient implementation (`CompositeHttpClient` and `HttpClientImpl`) is that it will _not_ introspect the URI to see if it is a full URI or not. Whatever is the URI specified will be passed as is to the server.**

This behavior with respect to scenario 1 above can be useful when the actual URI has the host and port information different than the server the request is sent on. As you had mentioned in one of our offline discussions and based on the [HTTP spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html) is a valid scenario for proxies. So, it does not make sense to disallow this scenario.

OTOH, this behavior w.r.t scenario 2 above can be rectified than what it exists today, by stripping the  scheme, host and port information from the URI in the `HttpClientRequest` before submitting to the client. IMO, this is the fix for issue #154.

Issue #172, opened 2014-07-03
Missing connection pool events from client built by RxContexts constructs


allenxwang:
Here is the JUnit test case that shows connection count is not incremented after a successful connection to www.google.com:

``` java
    @Test
    public void testConnectionPoolMetrics() {
        HttpClientBuilder<ByteBuf, ByteBuf> builder = RxContexts.newHttpClientBuilder("www.google.com", 80, RxContexts.DEFAULT_CORRELATOR);
        HttpClient<ByteBuf, ByteBuf> client = builder.withConnectionPoolLimitStrategy(new CompositePoolLimitDeterminationStrategy(
                new MaxConnectionsBasedStrategy(100), new MaxConnectionsBasedStrategy(100))).build();
        HttpClientListener listener = HttpClientListener.newHttpListener("default");
        client.subscribe(listener);
        HttpClientRequest<ByteBuf> request = HttpClientRequest.createGet("/");
        String result = client.submit(request).flatMap(new Func1<HttpClientResponse<ByteBuf>, Observable<ByteBuf>>(){
            @Override
            public Observable<ByteBuf> call(HttpClientResponse<ByteBuf> t1) {
                return t1.getContent();
            }
        }).map(new Func1<ByteBuf, String>(){
            @Override
            public String call(ByteBuf t1) {
                return t1.toString(Charset.defaultCharset());
            }
        }).toBlocking().single();
        System.out.println(listener.getConnectionCount());
        System.out.println(listener.getPoolAcquires());
    }
```

It prints out zero for both counts.

The poolAquires metric is wrong because of a typo in `HttpClientListener`:

``` java
         @Override
         protected void onPoolAcquireSuccess(long duration, TimeUnit timeUnit) {
            HttpClientListener.this.onPoolReleaseSuccess(duration, timeUnit); // wrong API called
         }
```

However, it seems that the connection count problem is caused by no listeners added to `ClientChannelFactoryImpl.eventsSubject`, after `RxContexts.withChannelFactory()` is called in RxContexts constructs.

NiteshKant:
Thanks @allenxwang for finding the issue.

The problem was that the client connection factory created by RxContexts was not having the reference to the eventsubject.This made the connection factory create a fresh subject.

PR #173, opened 2014-07-04
Fixes issue #172

NiteshKant:
The client connection factory created by RxContexts was not having the reference to the eventsubject.
This made the connection factory create a fresh subject.
Fixed the same and added the testcase for this.


NiteshKant:
The build failure was because I did not get the latest changes from the repo and there was a method signature change.

Issue #169, opened 2014-07-03
Remove ContentSource in favor of Observable for HttpClientRequest


NiteshKant:
`HttpClientRequest` does not use `Observable` source for its content. The intent of this was to retain the ability to put back pressure when the need arises.
Since, an `Observable` source did not initially provide a way to put back pressure, this approach was chosen initially.
RxJava is going to provide back pressure support with the regular `Observable` as discussed in [this issue](https://github.com/Netflix/RxJava/issues/1000). This is a more flexible solution (specifically when the source is actually an Observable, eg: for proxies) than the current `ContentSource` abstraction.
Since, the back pressure support in RxJava can be used with regular `Observables` it is safe now to replace `ContentSource` with `Observable` and later adopt back pressure in rxnetty.

PR #176, opened 2014-07-04
Fixes issue #169

NiteshKant:
- Modified HttpClientRequest to take an Observable for content instead of the earlier ContentSource.
- Retained the capability to use an Observable&lt;S&gt; where S is not the type of the HttpClientRequest. This mode requires passing a function that converts S to ByteBuf.
- Removed the RepeatableRequest functionality as it is not required. Re-subscribing to the content observable should replay the source.


allenxwang:
Looks good to me.

The only question I have is whether we can assume re-subscribing the content will replay the source. If the source is a PublishSubject, re-subscribing will only get the elements subsequent to the time of subscription. In that case, we might need to make use of ReplaySubject if we need to repeat the same request.

Of course, for majority of use cases the Observable is created using Observable.just() or from(), where re-subscribing will cause replay. We can also document the API to ask caller to provide the Observable which always replays when subscribing, if the request is deemed as "repeatable" by the caller.

NiteshKant:
The general contract of `Observable` is that it will replay the source on every fresh subscription unless it is a "hot observable" which is a live stream of events (eg: mouse events). 

I think people using rx would relate to this concept.

Issue #125, opened 2014-05-21
Websockets Proposal


NiteshKant:
### Scope

Netty supports websockets out of the box and hence it can be used within RxNetty with minimal effort (defining appropriate PipelineConfigurator to add websocket encoder/decoder). 
However, if the websockets implementation has to leverage RxNetty's HTTP abstraction (primarily of Client/Server request/response objects) few changes will be required in RxNetty.
In addition to these changes, following the philosophy of RxNetty, we would like to look into what can be done to make the usage of websockets in netty easier.
This proposal is aimed at providing these high level constructs in RxNetty to support websockets.
##### Client

Currently RxNetty's [HttpClientRequest](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/HttpClientRequest.java) does not support a stream of content, instead it has an iterative content source [ContentSource](https://github.com/Netflix/RxNetty/blob/master/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/ContentSource.java).
The reason for having this iterative contract instead of an Observable was the ability to apply backpressure from RxNetty in case, the client is too fast as compared to the underlying framework sending data over the wire.
However, it does impose a restriction for cases where there is a need of a stream of content, as is in this case, since a websocket connection is full duplex.

We have a few options here:
- Do away with the current ContentSource model in favor of an Observable which provides us the streaming support we need but sacrifices the backpressure capabilities.
- Create an extended WsClientRequest class which extends HttpClientRequest and adds an additional method to get content as an Observable. We can handle backpressure later when RxJava defines a backpressure contract as discussed [here](https://github.com/Netflix/RxJava/issues/1000)
- Expose the underlying netty's channel to the user so one can write the client side websocket frames anytime on the underlying channel. 
##### Server

There is no API change required for the server side as it does support streaming of request and response content.
### Contract

Provided we do the necessary changes to the API as described above, we should be able to provide the following 
constructs to handle websockets by client and server:
##### Client

```
Observable<HttpClientResponse<? extends WebSocketFrame> RxNetty.createWebSocketRequest(String uri);

```
##### Server

```
HttpServer<? extends WebSocketFrame, ? extends WebSocketFrame> RxNetty.createWebSocketServer(int port, RequestHandler<? extends WebSocketFrame, ? extends WebSocketFrame> handler)

```

[...]

PR #196, opened 2014-07-31
WebSockets - implementation proposal for issue #125

tbak:
This implementation is based solely on existing Netty's WebSockets implementation.

Client example:

``` java
WebSocketClient<WebSocketFrame> rxClient = RxNetty.newWebSocketClientBuilder("localhost", port)
            .withWebSocketURI("/websocket")
            .withWebSocketVersion(WebSocketVersion.V13)
            .build();

rxClient.connect().flatMap(
    new Func1<ObservableConnection<WebSocketFrame, WebSocketFrame>, Observable<Void>>() {
        ....
    });
```

Server example:

``` java
RxServer<WebSocketFrame, WebSocketFrame> server = RxNetty.
        newWebSocketServerBuilder(port, new ConnectionHandler<WebSocketFrame, WebSocketFrame>() {
            @Override
            public Observable<Void> handle(
                final ObservableConnection<WebSocketFrame, WebSocketFrame> newConnection) {
                    ....
                })
    }).build().start();
```


rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+        V08,
+
+        /**
+         * <a href="http://tools.ietf.org/html/rfc6455 ">RFC 6455</a>. This was originally <a href=
+         * "http://tools.ietf.org/html/draft-ietf-hybi-thewebsocketprotocol-17" >draft-ietf-hybi-thewebsocketprotocol-
+         * 17</a>
+         */
+        V13;
+
+        static WebSocketVersion toNetty(io.netty.handler.codec.http.websocketx.WebSocketVersion nettyVersion) {
```
NiteshKant: Should this be named fromNettyVersion() ?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public WebSocketClientBuilder<T> withMaxFramePayloadLength(int maxFramePayloadLength) {
+        this.maxFramePayloadLength = maxFramePayloadLength;
+        return this;
+    }
+
+    @Override
+    protected MetricEventsListener<? extends ClientMetricsEvent<? extends Enum>> newMetricsListener(MetricEventsListenerFactory factory, WebSocketClient<T> client) {
+        return null;
```
NiteshKant: Intentional?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientHandler.java
```diff
@@ -0,0 +1,98 @@
...
+import io.reactivex.netty.protocol.http.websocket.frame.NettyToRxFrameDecoder;
+import io.reactivex.netty.protocol.http.websocket.frame.RxToNettyFrameEncoder;
+
+/**
+ * {@link WebSocketClientHandler} orchestrates WebSocket handshake process and reconfigures
+ * pipeline after the handshake is complete.
+ *
+ * @author Tomasz Bak
+ */
+public class WebSocketClientHandler extends SimpleChannelInboundHandler<Object> {
```
NiteshKant: There are two main uses of `SimpleChannelInboundHandler`:

-- Be type aware (only accepts messages of a certain type)
-- Auto release ByteBuf

This code seems to be not using type awareness and the auto release to me sounds unnecessary as we always release the `ByteBuf` in `ObservableAdapter` after invoking the attached `Observer`'s `onNext()`.
So, looks like you can just use `ChannelInboundHandlerAdapter`

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
tbak: Yes. It is a mistake.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
tbak: Yes, I have mentioned in the pull request comment that metrics are not implemented yet. I think it is the last thing we should add after settling on a design.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientHandler.java
tbak: I agree. `ChannelInboundHandlerAdapter` is enough here.

tbak:
The current implementation of WebSockets is broken/limited at least on the server side, 
and actually the problem is of more generic nature.

For single HTTP server, we might wish to handle different URIs in different ways using different
protocols and stacks. For example:
1. plain HTTP request-response with ByteBuf
2. In pipeline encoder/decoder for certain URIs/object types
3. SSE for certain endpoints /a/b/c/see_source
4. WebSockets for certain endpoints /a/b/c/ws_endpoint
5. possibly other kinds with own specific pipeline configuration

It would be perfect if we could compose those in a flexible way. One way to do that might be
by extending existing HttpServer with a generic upgrade mechanism (to SSE, to WebSockets, etc) per
client connection. For example we could write:

``` java
RxNetty.newHttpServerBuilder(8080, new RequestHandler<ByteBuf, ByteBuf> {
    // Usual stuff here
}).upgradeTo(
        RxNetty.httpPipelineUpgrade("/a/b/c/my_special_object", new RequestHandler<MyObject, MyObject> {

        })
    )
  .upgradeTo(
        RxNetty.webSocketUpgrade("/a/b/c/ws_endpoint", new ObservableConnection<WebSocketFrame, WebSocketFrame> {
            // here we handle WebSockets stream
        })
    )
  .upgradeTo(
        RxNetty.sseUpgrade("/ab/c/sse_endpoint", new RequestHandler<ByteBuf, ServerSentEvent> {
            // SSE stream here
        })
  )
```

Exisiting methods like RxNetty.newWebSocketServerBuilder could be still left as a special cases.

gorzell:
I am meeting Nitseh today to talk about exactly this issue. You are welcome
to join.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClient.java
```diff
@@ -0,0 +1,77 @@
...
+    @SuppressWarnings("unchecked")
+    @Override
+    public Observable<ObservableConnection<O, I>> connect() {
+        return super.connect().lift(HANDSHAKE_OPERATOR);
+    }
+
+    static class HandshakeOperator<T extends WebSocketFrame> implements Operator<ObservableConnection<T, T>, ObservableConnection<T, T>> {
+        @Override
+        public Subscriber<ObservableConnection<T, T>> call(final Subscriber<? super ObservableConnection<T, T>> subscriber) {
+            return new Subscriber<ObservableConnection<T, T>>() {
```
NiteshKant: Won't you have to register the unsubscribe callback to unsubscribe when the passed subscriber unsubscribes?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+        return this;
+    }
+
+    public WebSocketClientBuilder<I, O> withWebSocketVersion(WebSocketVersion version) {
+        webSocketVersion = version;
+        return this;
+    }
+
+    public WebSocketClientBuilder<I, O> withMessageAggregator(boolean messageAggregation) {
```
NiteshKant: "withMessageAggregation" is a more appropriate name.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+        this.messageAggregation = messageAggregation;
+        return this;
+    }
+
+    public WebSocketClientBuilder<I, O> withSubprotocol(String subprotocol) {
+        this.subprotocol = subprotocol;
+        return this;
+    }
+
+    public WebSocketClientBuilder<I, O> withAllowExtensions(boolean allowExtensions) {
```
NiteshKant: "allowExtensions" may be a better name.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        if (!handshaker.isHandshakeComplete()) {
+            finishHandshake(ctx, (FullHttpResponse) msg, ch);
+        } else {
+            ctx.fireChannelRead(msg);
+        }
+    }
+
+    private void finishHandshake(ChannelHandlerContext ctx, FullHttpResponse msg, Channel ch) {
+        try {
+            handshaker.finishHandshake(ch, msg);
```
NiteshKant, line 91: Is this a synchronous call?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        } catch (WebSocketHandshakeException e) {
+            eventsSubject.onEvent(WebSocketClientMetricsEvent.HANDSHAKE_FAILURE, Clock.onEndMillis(handshakeStartTime));
+            handshakeFuture.setFailure(e);
```
NiteshKant, line 94: Do you want to use promise.tryXXX() methods instead of setXXX()? Otherwise, if this is already done, this will thrown an IllegalStateException.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketServer.java
```diff
@@ -0,0 +1,87 @@
...
+                WebSocketServerPipelineConfigurator<I, O> requiredConfigurator = (WebSocketServerPipelineConfigurator<I, O>) configurator;
+                requiredConfigurator.useMetricEventsSubject(eventsSubject);
+            }
+        }
+        if (!updatedSubject) {
+            throw new IllegalStateException("No server required configurator added.");
+        }
+    }
+
+    static class WrappedObservableConnectionHandler<I, O> implements ConnectionHandler<I, O> {
```
NiteshKant, line 60: Why is it named Wrapped?

NiteshKant, line 60: nevermind.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketServer.java
```diff
@@ -0,0 +1,87 @@
...
+
+        @Override
+        public Observable<Void> handle(final ObservableConnection<I, O> connection) {
+            ChannelHandlerContext ctx = connection.getChannelHandlerContext();
+            final ChannelPipeline p = ctx.channel().pipeline();
+            ChannelHandlerContext hctx = p.context(WebSocketServerHandler.class);
+            if (hctx != null) {
+                WebSocketServerHandler handler = p.get(WebSocketServerHandler.class);
+                final PublishSubject<Void> subject = PublishSubject.create();
+                handler.handshakeFuture().addListener(new ChannelFutureListener() {
```
NiteshKant: Is it worth seeing if we can delay `ObservableConnection` callback instead of delaying the Handler.onHandle() call?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClient.java
tbak: Good point. I will add this.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
tbak: Ok with me. I will rename it.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientBuilder.java
tbak: Ok

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientHandler.java
tbak, line 91: finishHandshake is called after we read HTTP upgrade message (we have FullHttpResponse object) and we can conclude the handshake.  It is a regular function call with immediate result.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketClientHandler.java
tbak, line 94: The code flow guarantees that the future will be set only once for both success and failure situations.
This could be violated if caller of `WebSocketClientHandler.handshakeFuture` manipulates the future object (it does not now, but some additional/custom handlers might). To prevent this I will replace the latter method with `WebSocketClientHandler.addHandshakeFinishedListener`.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/websocket/WebSocketServer.java
tbak: This means some work around ConnectionLifecycleHandler.
We could solve this problem in a generic way, to handle any connection upgrade, but for this we will need some extra API. Probably it should not be a public API, but more like SPI for building RxNetty extensions. So with this API I could say:
1. this is the first pipeline, it will call you back one work is done
2. this is second pipeline, apply it when callback is received
3. now it is time to hook-in the application connection handler

I would wait with modifying existing WebSockets implementation till we have this generic mechanism in place.

Issue #118, opened 2014-05-10
Tons of doc comment errors


headinthebox:
Trying to build with Java 8 and I get tons of erros in doc cemments :-(

```
RxNetty/rx-netty/src/main/java/io/reactivex/netty/protocol/http/sse/SSEInboundHandler.java:30: error: self-closing element not allowed
 * A handler to insert {@link ServerSentEventDecoder} at a proper position in the pipeline according to the protocol. <br/>
```

NiteshKant:
Thanks @headinthebox !
So ideally we will have to add these validations to the codestyle plugin we have so that we can catch these early.
@quidryan , any plans to add these to the gradle template?

@headinthebox If this is blocking you, you can add this ([source](http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html)):

```
if (JavaVersion.current().isJava8Compatible()) {
    allprojects {
      tasks.withType(Javadoc) {
        options.addStringOption('Xdoclint:none', '-quiet')
      }
    }
  }
```

temporarily to the build file.

headinthebox:
Thanks; don't understand why Java 8 is so picky about this.

NiteshKant:
@headinthebox yeah "self-closing element not allowed" sound insane as it is a valid HTML.

quidryan:
The plan is to support RxJava going to a new github organization via a new Gradle plugin. I'd then port that plugin back for NetflixOSS, so give it a few weeks. I just put this change into the RxJava version: https://github.com/nebula-plugins/gradle-rxjava-project-plugin/commit/d755bb771ee6e94b2c36c5fac97f0ac8e24439b7

NiteshKant:
As per PR #201 doclint has been disabled for java8.

quidryan:
Nothings going to happen on gradle-template.  We're redesigning it, you can see an incarnation of it in https://github.com/nebula-plugins/gradle-rxjava-project-plugin. That already disables the lint check in Java 8.

gorzell:
Fixing the documentation is still a better option than disabling it by
default ;-).

quidryan:
I guess on a new project that's true. Existing projects haven't had it on
for literally a decade. It's a pretty big change to slip into a JVM
release. Why turn it on after all these years?

gorzell:
Probably to push people in the "right" direction.

PR #201, opened 2014-08-09
Disabling doclint for Java 8

diptanu:
RxNetty doesn't build on Java 8 because some of the java docs are not complaint to W3C standards. Please see http://stackoverflow.com/questions/22528767/jdk8-and-javadoc-has-become-very-strict for reference.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,9 @@ allprojects {
         mavenLocal()
         mavenCentral() // maven { url: 'http://jcenter.bintray.com' }
     }
+    tasks.withType(Javadoc) {
+        options.addStringOption('Xdoclint:none', '-quiet')
+    }
 }
 
 apply from: file('gradle/convention.gradle')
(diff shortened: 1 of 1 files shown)


diptanu:
Looks like the builds are failing because of some test failure?

tbak:
I would leave the doclint on, and fix the non-compliant javadocs. We did this already for Ribbon.

NiteshKant:
@tbak, since this is blocking anyone building on jdk 8, should we merge this and then remove the disable of doclint when our javadocs are fixed?

I would actually have this disabled since we are not actively building in java 8 & there isn't a way to make sure we don't break this unknowingly.

There is a related issue #118 we can close that when we merge this.

tbak:
@NiteshKant Lets do this as a workaround, but I think it makes sense to do regular builds with Java8, specially that many people depend on it.

PR #204, opened 2014-08-12
Add a generic Handler interface

gorzell:
Add a generic Handler interface and change the RequestHandler to be a child interface of the generic one.


NiteshKant:
Thanks @gorzell !

PR #207, opened 2014-08-18
Fixes #206

### LINKED ISSUE #206: HttpClientResponse.getContent() will loose data if not eagerly subscribed
`HttpClientResponse` expects users to eagerly subscribe to the content when they receive the `HttpClientResponse`. This means that one can not do this:

``` java
        RxNetty.createHttpGet("http://www.google.com")
                    .toBlocking().toFuture().get(1, TimeUnit.MINUTES).getContent()
                    .forEach(System.out::println);
```

Instead the following has to be used:

``` java
        RxNetty.createHttpGet("http://www.google.com")
               .flatMap(HttpClientResponse::getContent)
               .map(byteBuf -> {
                   System.out.println(byteBuf.toString(Charset.defaultCharset()));
                   return null;
               })
               .filter(aVoid -> false).toBlocking().lastOrDefault(null);
```

The reason being that RxNetty starts publishing the content to a `PublishSubject` immediately after the `HttpClientResponse` `onNext()` call. 
In the first case, the subscription to content happens after the first `Observable` completes which is when the content completes. In the other case, the subscription happens in the Rx chain of processing `HttpClientResponse` via the `flatMap` and hence the content is correctly subscribed.

The reason why this was done was to eliminate the need of caching the content to enable lazy subscription to the content `Observable`. However, this is pretty limiting because if the code happens to apply an operator that waits for `onComplete` of the source (returned by `RxNetty.createHttpGet()` in this example) like `single()` or `.toBlocking().toFuture().get()` in this case the code silently ignores the content.
### Solution

We need two things to be fixed:
1. The main `Observable`, returned by `RxNetty.createHttpXXX()` methods and returned by `HttpClient.submit()` must _not_ complete after the content is completed. It should complete when the `HttpClientResponse` is delivered (there can every only be 1 of these).
2. The content `Observable` must cache the content and allow lazy subscription.
### Drawbacks

The main drawback of doing this is that the content has to be cached, what happens if the user never subscribes to the content?

There are two things that should be done, viz.,
1. Have a method `ignoreContent()` on `HttpClientResponse` that indicates that the user will never subscribe to content. This will not cache the content in the subject.
2. Detect such leaks (if `ignoreContent()` is not called and content is not subscribed) and do verbose logging.
### How about `HttpServerRequest`?

Since `RequestHandler` gives a callback on `HttpServerRequest` and not an `Observable<HttpServerRequest>` this does not seem to be a big issue however it has the same problem i.e. if the content is not eagerly subscribed (inside `RequestHandler.handle()`) it will loose data.

NiteshKant:
Making HttpServerRequest.getContent() and HttpClientResponse.getContent() lazily subscribable.

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
@@ -68,11 +68,6 @@ import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
  */
 public final class UnicastContentSubject<T> extends Subject<T, T> {
 
-    private static final IllegalStateException ALREADY_DISPOSED_EXCEPTION =
-            new IllegalStateException("Content stream is already disposed.");
-    public static final IllegalStateException MULTIPLE_SUBSCRIBERS_EXCEPTIONS =
-            new IllegalStateException("Content can only have one subscription. Use Observable.publish() if you want to multicast.");
-
     private final State<T> state;
     private volatile Observable<Long> timeoutScheduler;
 
@@ -252,9 +247,9 @@ public final class UnicastContentSubject<T> extends Subject<T, T> {
                     }
                 }));
             } else if(State.STATES.SUBSCRIBED.ordinal() == state.state) {
-                subscriber.onError(MULTIPLE_SUBSCRIBERS_EXCEPTIONS);
+                subscriber.onError(new IllegalStateException("Content can only have one subscription. Use Observable.publish() if you want to multicast."));
             } else if(State.STATES.DISPOSED.ordinal() == state.state) {
-                subscriber.onError(ALREADY_DISPOSED_EXCEPTION);
+                subscriber.onError(new IllegalStateException("Content stream is already disposed."));
             }
         }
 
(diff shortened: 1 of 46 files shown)


NiteshKant:
The last failure seems to be because of Cloudbees' local maven repo corruption.

I am disabling local repo for RxNetty builds, that seems to be the only solution for this.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * This implementation is inspired by
+ * <a href="https://github.com/Netflix/RxJava/blob/master/rxjava-core/src/main/java/rx/internal/operators/BufferUntilSubscriber.java">RxJava's BufferUntilSubscriber</a>
+ * @author Nitesh Kant
+ */
+public final class UnicastContentSubject<T> extends Subject<T, T> {
+
+    private static final IllegalStateException ALREADY_DISPOSED_EXCEPTION =
+            new IllegalStateException("Content stream is already disposed.");
+    public static final IllegalStateException MULTIPLE_SUBSCRIBERS_EXCEPTIONS =
```
tbak: It saves CPU/memory to have it pre-created, but stack trace is misleading to users.

NiteshKant: Agreed. Removed the pre-created exceptions.

Issue #209, opened 2014-08-19
Deprecate all SSE classes in io.reactivex.netty.protocol.text.sse


NiteshKant:
In relation to issue #205 we need to first deprecate these classes in the io.reactivex.netty.protocol.text.sse package so that we can delete some and move the other to the io.reactivex.netty.protocol.http.sse package in the next release.

gorzell:
I have this mostly done.

What I did was move the existing classes and replaced them with proxies that extend the originals.  These proxies are marked as deprecated.

NiteshKant:
@gorzell sounds good, I will wait for your PR then. In case I need to release 0.3.12 sooner, will punt this to 0.3.13

gorzell:
I'll try to wrap this up today.  It has been a bit tricky because some tests fail sporadically for me and the Context tests still fail every time.

Issue #205, opened 2014-08-15
Move SSE related class to the http package


gorzell:
Right now things related to SSE are in a text package, which doesn't make a lot of sense since SSE is an http dependent protocol.  Also websockets are a sub package of HTTP so this change would make SSE consistent with that as well.

NiteshKant:
:+1: We should not have SSE over text option.

@jstnbckr @g9yuayon is mantis using the SSE over text codecs? Is it ok to remove them and have SSE only in HTTP?

If it's ok with everyone, I will deprecate all the existing SSE over text codecs and remove them after 1 release cycle.

g9yuayon:
As far as I know, we're not using text codecs. That said, moving to HTTP
makes more sense as SSE is part of HTML 5 spec and is mainly used for HTTP.

NiteshKant:
Thanks @g9yuayon 

I will deprecate the text codecs in 0.3.12 and we can remove them in 0.3.13 and move the required classed under HTTP package.

@gorzell I hear that you will work on this :)

PR #210, opened 2014-08-19
Sse migration from text to http package

gorzell:
This should address most of #205 and does address #209.

As mentioned in #209.  I moved the original classes and then created subclasses with the same name in the text.sse package.  These sub classes are marked as deprecated and will be removed, but they should provide backwards compatibility via casting to their super classes for anyone that happens to be using them in explicitly in their code.


rx-netty/src/main/java/io/reactivex/netty/protocol/http/sse/SSEClientPipelineConfigurator.java
```diff
@@ -0,0 +1,44 @@
...
+ * This will convert {@link ByteBuf} objects to {@link ServerSentEvent}. So, if the client is an HTTP client, then you would
+ * have to use {@link SseOverHttpClientPipelineConfigurator} instead.
+ *
+ * @param <W> The request type for the client pipeline.
+ *
+ * @see ServerSentEventDecoder
+ *
+ * @author Nitesh Kant
+ */
+public class SSEClientPipelineConfigurator<W> implements PipelineConfigurator<ServerSentEvent, W> {
```
NiteshKant, line 36: Is there value in keeping these classes since SSE can only be used in HTTP?

HTTP handlers use these only for referring to the handler instance.

If no, then we can just keep them deprecated & have a singleton `SSEInboundHandler` instance in the HTTP handler.

Same for client.

NiteshKant:
Cloudbees PR verification job isn't running after the move to ReactiveX org. We are fixing the build, till then I ran this PR locally to verify that tests passed.

@gorzell I will wait for your reply on my comment and then I can merge this PR.

gorzell:
Should be later tonight that I can look at it.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/sse/SSEClientPipelineConfigurator.java
gorzell, line 36: Probably not.  I'll try and get it patched tonight.

gorzell, line 36: After looking more closely I am not quite sure I understand where/how you want this moved/removed.  I suppose you could have it extend `HttpServerPipelineConfigurator`.  But that may not be such a good idea as you don't know whether a user might want to put their own HTTP related handers between the base HTTP handler and where they do the decoding.

I do think however that the `SseOverHttpClientPipelineConfigurator` classes should be removed or moved into factory methods for default pipeline configurations.

NiteshKant, line 36: My point was that we do not need both `SseOverHttpClientPipelineConfigurator` and `SSEClientPipelineConfigurator` (same for server).

So, we don't need to add `SSEClientPipelineConfigurator` to http package. Just fold it into `SseOverHttpClientPipelineConfigurator`

Alternatively, deprecate `SseOverHttpClientPipelineConfigurator` and just have `SSEClientPipelineConfigurator` in http package containing what it takes to run SSE.

gorzell, line 36: I would prefer the latter, and will add the deprecation flags.

Issue #31, opened 2014-01-29
Simple URI mapping


g9yuayon:
We'd still need an simple URI mapping that dispatches different handlers to different URI patterns. Mantis already has a few use cases for this function. I'd imagine that URI  mapping is the first thing that most users will need, it would make sense to put it in RxNetty.

NiteshKant:
In the simplest form I guess what you are saying is provide a Map of uri regex Vs Action1 implementations, that the Action1 instance passed to RxServer.start() uses, rite?

https://github.com/Netflix/RxNetty/blob/master/src/main/java/io/reactivex/netty/server/RxServer.java#L60

NiteshKant:
@g9yuayon I would not have time to work on this in the short term, let me know if you want to work on this. In that case, I would recommend you to wait till I am done with the latest changes in APIs.

g9yuayon:
No worries, @NiteshKant. In that case I'll work on it.

NiteshKant:
This will be covered by #286. Closing this as duplicate.

PR #211, opened 2014-08-19
HTTP file server

elandau:
Initial implementation for serving static content from RxNetty based on https://github.com/netty/netty/blob/master/example/src/main/java/io/netty/example/http/file/HttpStaticFileServerHandler.java.  

Additional features include,
1.  Support for single and multiple file sources
2.  zero-copy serving of files


gorzell:
yay!

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        
+        try {
+            final String uri = request.getUri();
+            URL resource = resolver.getUrl(uri);
+            if (resource == null) {
+                return FileResponses.sendError(response, HttpResponseStatus.NOT_FOUND);
+            }
+            
+            File file = new File(resource.toURI());
```
tbak: What if the file found on classpath is in jar not on the file system?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,100 @@
...
+    public Observable<Void> handle(
+            HttpServerRequest<ByteBuf> request,
+            HttpServerResponse<ByteBuf> response) {
+        
+        if (!request.getHttpMethod().equals(HttpMethod.GET)) {
+            return Observable.empty();
+        }
+        
+        try {
+            final String uri = request.getUri();
```
tbak: Should not we have UriResolver? URL class is broken, and we anyway convert it to URI here.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/OwnableRequestHandler.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Encapsulate a request handler and the URLResolver.  Used in {@link MultipleFileRequesthandler}
+ * 
+ * @author elandau
+ *
+ * @param <I>
+ * @param <O>
+ */
+public class OwnableRequestHandler<I, O> implements RequestHandler<I, O> {
+
```
tbak: This is useful not only for static content handling, so we could put it into server package.
And maybe find a better name, like RequestHandlerRouter or RequestHandlerDispatcher? 

Alternative implementation could follow the chain of responsibility pattern. Would be more generic, and applicable to other cases as well.

rx-netty/src/main/resources/META-INF/mime.types
```diff
@@ -0,0 +1,1588 @@
...
+# here to guide configurations toward known types and to make it easier to
+# identify "new" types.  File extensions are also commonly used to indicate
+# content languages and encodings, so choose them carefully.
+#
+# Internet media types should be registered as described in RFC 4288.
+# The registry is at <http://www.iana.org/assignments/media-types/>.
+#
+# MIME type (lowercased)			Extensions
+# ============================================	==========
+# application/1d-interleaved-parityfec
```
tbak: Is there a reason that we have some mime types filtered out?
If yes, why not remove them from this file.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/OwnableRequestHandler.java
elandau: +1 
A more generic solution would be great

rx-netty/src/main/resources/META-INF/mime.types
elandau: Good question.  We may want to seriously trim the supported types in this file and make sure there is a mechanism to provide addition mime types.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
elandau: I'll make the change.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
elandau: This will handle files in jars as well

rx-netty/src/main/java/io/reactivex/netty/channel/DefaultChannelWriter.java
```diff
@@ -102,6 +106,28 @@ public void writeString(String msg) {
         write(msg, new StringTransformer());
         return flush();
     }
+    
+    @Override
+    public Observable<Void> writeFileRegion(FileRegion region) {
+        final ChannelFuture future = writeOnChannel(region);
```
NiteshKant: We should not use writeOnChannel() here as that adds the future of write to unflushed futures. This write does not need a flush.

Instead you can just use 

``` java
ChannelFuture writeFuture = getChannel().write(region);
```

rx-netty/src/main/java/io/reactivex/netty/channel/ChannelWriter.java
```diff
@@ -36,6 +37,8 @@
 
     void writeString(String msg);
 
+    Observable<Void> writeFileRegion(FileRegion region);
```
NiteshKant: Should we instead return `Observable<Long>` and underneath use a `ProgressivePromise` to provide the status of the write in terms of bytes written?

The code then would be:

``` java
        final ChannelProgressivePromise progressivePromise = ctx.newProgressivePromise();
        getChannel().write(region, progressivePromise);
        return Observable.create(new Observable.OnSubscribe<Long>() {
            @Override
            public void call(final Subscriber<? super Long> subscriber) {
                progressivePromise.addListener(new ChannelProgressiveFutureListener() {
                    @Override
                    public void operationProgressed(ChannelProgressiveFuture future, long progress, long total)
                            throws Exception {
                        subscriber.onNext(progress); // sending progress, total can be calculated by the subscriber.
                    }

                    @Override
                    public void operationComplete(ChannelProgressiveFuture future) throws Exception {
                        if (future.isSuccess()) {
                            subscriber.onCompleted();
                        } else {
                            subscriber.onError(future.cause());
                        }
                    }
                });
            }
        });
```

rx-netty/src/main/java/io/reactivex/netty/channel/DefaultChannelWriter.java
elandau: This doesn't work.  It seems like the write does need to be flushed after all.

rx-netty/src/main/java/io/reactivex/netty/channel/ChannelWriter.java
elandau: Makes sense but how will this Observable<Long> be consumed?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/OwnableRequestHandler.java
NiteshKant: Are we talking about a simple URI router much like discussed in issue #31 ?

If yes, I see it much like karyon's [SimpleUriRouter](https://github.com/Netflix/karyon/blob/master/karyon-core/src/main/java/com/netflix/karyon/transport/http/SimpleUriRouter.java)

I am for creating this utility to compose multiple `RequestHandler` as handling multiple URIs.

elandau: I think this use case is different from what's implemented in SimpleUriRouter.  In SimpleUriRouter each uri prefix of regex is mapped to exactly one RequestHandler.  Here we can specify a cascading set of file sources that are inspected in a predefined order.  So, depending on which source contains the file the same URI may be end up using a different RequestHandler.  It would be equivalent to SimpleUriRouter mapping a URI to multiple RequestHandlers.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Base implementation for serving local files.  Resolving the request URI to
+ * a local file URI is deferred to the subclass.
+ * 
+ * @author elandau
+ *
+ */
+public class FileRequestHandler implements RequestHandler<ByteBuf, ByteBuf> {
+    public static final String IF_MODIFIED_SINCE = "If-Modified-Since";
```
NiteshKant: You can use netty's HTTP header name: `io.netty.handler.codec.http.HttpHeaders.Names.IF_MODIFIED_SINCE`

rx-netty/src/main/resources/META-INF/mime.types
```diff
@@ -0,0 +1,1588 @@
+# This file maps Internet media types to unique file extension(s).
```
NiteshKant, line 1: Where is this file used?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/OwnableRequestHandler.java
NiteshKant: Sure but the URI -> `RequestHandler` mapping can also map to a composed `RequestHandler`, correct?

To concretize what I am saying:

``` java
public class SimpleUriRouter<I, O> implements RequestHandler<I, O> {

    public SimpleUriRouter<I, O> withRoute(String uri, RequestHandler<I, O> route) {
        // store the route in a linked hashmap to guarantee route matching in the order of addition
        return this;
    }

    @Override
    public Observable<Void> handle(HttpServerRequest<I> request, HttpServerResponse<O> response) {
        return Observable.empty(); //TODO: Run through the routes and choose whichever matches first
    }

    public static void main(String[] args) {
        SimpleUriRouter<ByteBuf, ByteBuf> router = new SimpleUriRouter<>();
        router.withRoute("/a", (request, response) -> Observable.empty());

        SimpleUriRouter<ByteBuf, ByteBuf> complexRoute = new SimpleUriRouter<>(); 
        complexRoute.withRoute("/b/c", (request, response) -> Observable.empty());
        complexRoute.withRoute("/b/d", (request, response) -> Observable.empty());

        router.withRoute("/b", complexRoute);
    }
}
```

tbak: If we replace first argument in `withRoute` with a predicate function, we will get more flexibility. We could provide a few standard predicates for string prefix, regexp.

elandau: But the above example still has a 1-1 relationship between path and handler.  The functionality I've implemented would be more along the lines of 

``` java
router.withRoute("/b", ImmutableList.of(new FileSystemBasedRouter("/var/files"), new ClassPathBaseRouter(), new RemoteUrlBasedRouter("http://someotherfileserver"));
```

The expected behavior would be that for any uri starting with "/b" we'd first look for the file in the folder /var/files.  If not found then look for the file in the class path.  If still not found then look for the file on a remote server.

rx-netty/src/main/resources/META-INF/mime.types
elandau, line 1: This is used in FileResponses.setContentTypeHeader().

``` java
    public static void setContentTypeHeader(HttpServerResponse<ByteBuf> response, File file) {
        MimetypesFileTypeMap mimeTypesMap = new MimetypesFileTypeMap();
        response.getHeaders().set(HttpHeaders.CONTENT_TYPE, mimeTypesMap.getContentType(file.getPath()));
    }
```

See documentation here, http://docs.oracle.com/javase/7/docs/api/javax/activation/MimetypesFileTypeMap.html

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/ClassPathURLResolver.java
```diff
@@ -0,0 +1,28 @@
...
+    public ClassPathURLResolver() {
+        this(DEFAULT_PATH_PREFIX);
+    }
+    
+    public ClassPathURLResolver(String prefix) {
+        this.prefix = prefix;
+    }
+     
+    @Override
+    public URL getUrl(String path) {
```
gorzell: Might be worth comment that this can return null.

gorzell:
No unit tests at all?!?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/ClassPathURIResolver.java
```diff
@@ -0,0 +1,37 @@
...
+import java.net.URL;
+
+/**
+ * Resolve the URL for a request URI from the classpath
+ * 
+ * @author elandau
+ *
+ */
+public class ClassPathURIResolver implements URIResolver {
+    private static final String DEFAULT_PATH_PREFIX = "WEB-INF";
```
gorzell: Should you check for and enforce that this ends with a "/"?  You add a file separator in the LocalDirectoryResolver.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/LocalDirectoryURIResolver.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+    
+    public LocalDirectoryURIResolver(String prefix) {
+        this.prefix = prefix;
+    }
+     
+    @Override
+    public URI getUri(String path) {
+        try {
+            return new URI("file:///" + prefix + path);
```
gorzell: There is a security issue here, there is nothing to prevent the caller from navigating the filesystem.  This may also exist for the classpath, but I am having a harder time proving that.

```
val uri2 = new URI(s"file:///$prefix/../../etc/hosts")
uri2: java.net.URI = file:////Users/gorzell/../../etc/hosts

scala> import java.io.File
import java.io.File

scala> new File(uri2).exists
res10: Boolean = true
```

Some unit tests with these kinds of cases would really help.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,98 @@
...
+                }
+            }
+            
+            response.setStatus(HttpResponseStatus.OK);
+            response.getHeaders().setContentLength(fileLength);
+            FileResponses.setContentTypeHeader(response, file);
+            FileResponses.setDateAndCacheHeaders(response, file);
+            response.writeFileRegion(new DefaultFileRegion(raf.getChannel(), 0, fileLength));
+            return response.close();
+//            if (HttpHeaderUtil.isKeepAlive(request)) {
```
gorzell: Probably worth cleaning this up to avoid confusion.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/OwnableRequestHandler.java
gorzell: I think this might be another place where it would be more clear if we had a good mechanism for building chains of handlers.  I think that is in essence what @elandau is trying to setup here.  But what each handler in the chain does with the request should be arbitrary.

This kind of gets back to the question of what is the difference between a chain of handlers and a pipeline and how we make both easy to configure and combine.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/URIResolver.java
```diff
@@ -0,0 +1,18 @@
...
+ * Plugable policy to resolve a request URI to a local URL
+ * 
+ * @author elandau
+ */
+public interface URIResolver {
+    /**
+     * Resolve the path URI to a local URL 
+     * 
+     * @param uri
+     * @return A local URL or null if not found
```
gorzell: This isn't really accurate (and has a typo).  The LocalFile implementation doesn't check that the file is "found" or exists at all.  Short of a totally malformed string, not much in the way of sanity checks are promised.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/UrlUtils.java
```diff
@@ -0,0 +1,38 @@
+package io.reactivex.netty.protocol.http.server.file;
+
+import java.io.File;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.util.regex.Pattern;
+
+public class UrlUtils {
```
gorzell: This is at least a start, I don't see where it is ever used though.  Is the find feature in Chrome failing me?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/URIResolver.java
```diff
@@ -0,0 +1,18 @@
+package io.reactivex.netty.protocol.http.server.file;
+
+import java.net.URI;
+
+/**
+ * Plugable policy to resolve a request URI to a local URL
+ * 
+ * @author elandau
+ */
+public interface URIResolver {
```
gorzell: I'd suggest an abstract base class that does some of the basics like sanitizing the input (unless I missed where that is happening) which could then be extended by the two existing impls you have.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/UrlUtils.java
elandau: Using this somehow got dropped from the original code.  It's back in there now.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/URIResolver.java
elandau: The contract should be null if not found.  I fixed LocalDirectoryURIResolver to check for existence.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/URIResolver.java
elandau: Sanitize now happens early on so we can return a 403

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/LocalDirectoryURIResolver.java
elandau: done

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
elandau: done

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/ClassPathURIResolver.java
elandau: done

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/LocalDirectoryRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    protected URI resolveUri(String path) {
+        try {
+            URI uri = new URI("file:///" + prefix + path);
+            if (Files.notExists(Paths.get(uri), LinkOption.NOFOLLOW_LINKS)) {
+                return null;
+            }
+            return uri;
+        } catch (URISyntaxException e) {
+            return null;
```
NiteshKant, line 43: Log the exception?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/WebappFileRequestHandler.java
```diff
@@ -0,0 +1,7 @@
+package io.reactivex.netty.protocol.http.server.file;
+
+public class WebappFileRequestHandler extends ClassPathFileRequestHandler {
```
NiteshKant, line 3: We don't really have much webapp support perse :)

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/LocalDirectoryRequestHandler.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+    
+    public LocalDirectoryRequestHandler(String prefix) {
+        this.prefix = prefix;
+    }
+     
+    @Override
+    protected URI resolveUri(String path) {
+        try {
+            URI uri = new URI("file:///" + prefix + path);
```
NiteshKant: May be you can pre-create `"file:///" + prefix` string to avoid one more string creation?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/LocalDirectoryRequestHandler.java
```diff
@@ -0,0 +1,40 @@
+package io.reactivex.netty.protocol.http.server.file;
+
+import io.netty.util.internal.SystemPropertyUtil;
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
```
NiteshKant: Looks like this is a java 7 API, we are java 6 compatible.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,111 @@
...
+ * 
+ * @author elandau
+ *
+ */
+public abstract class FileRequestHandler extends AbstractFileRequestHandler {
+    @Override
+    public Observable<Void> handle(HttpServerRequest<ByteBuf> request, HttpServerResponse<ByteBuf> response) {
+        // We don't support GET.  
+        if (!request.getHttpMethod().equals(HttpMethod.GET)) {
+            return Observable.error(new HttpError(HttpResponseStatus.NOT_FOUND));
```
NiteshKant: If GET is not supported, then should the response code be `405` => Method Not Allowed?

BTW, why is GET not supported?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,111 @@
...
+                SimpleDateFormat dateFormatter = new SimpleDateFormat(HTTP_DATE_FORMAT, Locale.US);
+                Date ifModifiedSinceDate = dateFormatter.parse(ifModifiedSince);
+
+                // Only compare up to the second because the datetime format we send to the client
+                // does not have milliseconds
+                long ifModifiedSinceDateSeconds = ifModifiedSinceDate.getTime() / 1000;
+                long fileLastModifiedSeconds = file.lastModified() / 1000;
+                if (ifModifiedSinceDateSeconds == fileLastModifiedSeconds) {
+                    response.setStatus(HttpResponseStatus.NOT_MODIFIED);
+                    setDateHeader(response);
```
NiteshKant: Is there a value in passing the `SimpleDateFormat` instance to this method? This will also aid method reuse from `setDateAndCacheHeaders`

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            response.setStatus(HttpResponseStatus.OK);
+            response.getHeaders().setContentLength(fileLength);
+            setContentTypeHeader(response, file);
+            setDateAndCacheHeaders(response, file);
+            response.writeFileRegion(new DefaultFileRegion(raf.getChannel(), 0, fileLength));
+            
+            // TODO: Handle keep alive headers
+            return response.close().finallyDo(closeFileAction(raf));
+        }
+        catch (Exception e) {
```
NiteshKant: You do not have to add this error handling logic as `HttpServer` does the handling like this by default.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+            
+            response.setStatus(HttpResponseStatus.OK);
+            response.getHeaders().setContentLength(fileLength);
+            setContentTypeHeader(response, file);
+            setDateAndCacheHeaders(response, file);
+            response.writeFileRegion(new DefaultFileRegion(raf.getChannel(), 0, fileLength));
+            
+            // TODO: Handle keep alive headers
```
NiteshKant: Is this a TODO we can't resolve now?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
elandau: Only GET is supported.  I'll change the response to 405.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
elandau: Hmmm... but I have to catch the checked exceptions somewhere.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,111 @@
...
+                    setDateHeader(response);
+                    return response.close().finallyDo(closeFileAction(raf));
+                }
+            }
+            
+            response.setStatus(HttpResponseStatus.OK);
+            response.getHeaders().setContentLength(fileLength);
+            setContentTypeHeader(response, file);
+            setDateAndCacheHeaders(response, file);
+            response.writeFileRegion(new DefaultFileRegion(raf.getChannel(), 0, fileLength));
```
NiteshKant, line 126: The [netty example](https://github.com/netty/netty/blob/master/example/src/main/java/io/netty/example/file/FileServerHandler.java) suggests that we can not use this for HTTPS. Should we be handling SSL case?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            
+            // TODO: Handle keep alive headers
+            return response.close().finallyDo(closeFileAction(raf));
```
NiteshKant: By reading code it appears that netty will close the file after writing it on the channel. 

Specifically, the method [DefaultFileRegion.deallocate()](https://github.com/netty/netty/blob/master/example/src/main/java/io/netty/example/file/FileServerHandler.java) will be called by [ChannelOutboundBuffer](https://github.com/netty/netty/blob/master/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java#L246) post transfer. So, I think we can safely remove this close handler.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/RequestHandlerWithErrorMapper.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Decorator for a {@link RequestHandler} with an accompanying {@link ErrorResponseGenerator}.
+ * 
+ * @author elandau
+ *
+ * @param <I>
+ * @param <O>
+ */
+public class RequestHandlerWithErrorMapper<I, O> implements RequestHandler<I, O> {
+    private Func1<Throwable, ErrorResponseGenerator<O>> errorMapper;
```
NiteshKant, line 15: Can this just be `Func2<Throwable, HttpServerResponse<I>, Observable<Void>>` or just `Func2<Throwable, HttpServerResponse<I>, Void>`?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/RequestHandlerWithErrorMapper.java
```diff
@@ -0,0 +1,46 @@
...
+            RequestHandler<I, O> handler,
+            Func1<Throwable, ErrorResponseGenerator<O>> errorMapper) {
+        this.handler = handler;
+        this.errorMapper = errorMapper;
+    }
+
+    @Override
+    public Observable<Void> handle(HttpServerRequest<I> request, final HttpServerResponse<O> response) {
+        return handler.handle(request, response)
+                .onErrorResumeNext(new Func1<Throwable, Observable<Void>>() {
```
NiteshKant, line 32: Does it make sense to just do a `doOnError()` so that you can avoid creating and managing an `Observable`. Specifically since the `errorMapper` are sync (returns void), it should just be fine?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/file/FileRequestHandler.java
NiteshKant: Oh there are checked exceptions thrown by this code? Since, it was effectively the entire method in a try-catch and rxnetty APIs not throwing an exception I thought it is just a safe-guard.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/RequestHandlerWithErrorMapper.java
elandau, line 15: Don't we want to promote reuse of ErrorResponseGenerator as the contract for generating error responses?  Why introduce another contract.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/RequestHandlerWithErrorMapper.java
elandau, line 32: Won't doOnError still propagate the error to to the subscriber of Observable<Void>.  The idea here is to write the error response and not propagate the error to the container.

PR #224, opened 2014-09-05
Fixes #223 (Client connections receiving SSE responses should never be pooled & reused)

### LINKED ISSUE #223: Client connections receiving SSE responses should never be pooled & reused
Server sent events is an endless stream and hence apart from closing the connection there is no way for the client to stop receiving data.
In RxNetty the way for a client to close the connection is to unsubscribe from the content `Observable` in the `HttpClientResponse`. 
In cases when the `HttpClient` is using a connection pool (default client does use pooling) and it receives an SSE response, the unsubscribe from the content `Observable` does not result in closing the underlying physical connection. This results in a connection which is still receiving data but no one is reading from it and when another request reuses this connection to send another HTTP request, it is invalid.

Currently, in order to get past this issue it is advised for people to specifically disable connection pooling for the clients that may process an SSE request by using `HttpClientBuilder.withNoConnectionPooling()`. With this change, anytime an SSE response is received on a connection it will be discarded whenever the client unsubscribes from the response stream.

NiteshKant:
Channel attribute "rxnetty_http_discard_connection" is used by `ObservableConnection` to discard a connection on close. SSE Inbound handler sets this attribute as a connection on which SSE response has started should never be pooled.

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/protocol/http/sse/SSEInboundHandler.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/protocol/http/sse/SSEInboundHandler.java
@@ -24,6 +24,7 @@ import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpHeaders;
 import io.netty.handler.codec.http.HttpResponse;
 import io.netty.handler.codec.http.LastHttpContent;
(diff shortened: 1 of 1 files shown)


NiteshKant:
Closing this PR as it requires #225 to be fixed.

PR #234, opened 2014-09-16
Updated so that the SSLEngines are lazy loaded.

### LINKED ISSUE #233: Using DefaultFactories.TRUST_ALL causes SelfSignedSSLEngineFactory to Error on Certain Systems.
On systems that don't have OpenJdkSelfSignedCertGenerator or BouncyCastleSelfSignedCertGenerator available to them (for example Android), it will error out when using DefaultFactories.TRUST_ALL, even though those are requirements for DefaultFactories.SELF_SIGNED.

This could be solved by having the two engines SELF_SIGNED and TRUST_ALL be lazy loaded, so that they are only instantiated when needed. Let me know how you would like to proceed with this and I can do the work and provide a PR.

Also a side note, shouldn't these by final?

AndrewReitz:
TRUST_ALL and SELF_SIGNED where both loaded at the same time. In environments where only TRUST_ALL was needed and OpenJdkSelfSignedCertGenerator or BouncyCastleSelfSignedCertGenerator where not available. SELF_SIGNED would throw an exception forcing the user to create their own SSLEngineFactory for this type.

This is a breaking change for users that were using SELF_SIGNED or TRUST_ALL.

Fixes #233

### DIFF
--- a/rx-netty-examples/src/main/java/io/reactivex/netty/examples/http/ssl/README.md
+++ b/rx-netty-examples/src/main/java/io/reactivex/netty/examples/http/ssl/README.md
@@ -30,7 +30,7 @@ Here is the snippet from [SslHelloWordClient](SslHelloWorldClient.java):
 ```java
 public HttpResponseStatus sendHelloRequest() throws Exception {
     HttpClient<ByteBuf, ByteBuf> rxClient = RxNetty.<ByteBuf, ByteBuf>newHttpClientBuilder("localhost", port)
-            .withSslEngineFactory(DefaultFactories.TRUST_ALL)
+            .withSslEngineFactory(DefaultFactories.trustAll())
             .build();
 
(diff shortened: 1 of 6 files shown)


NiteshKant:
Thanks @pieces029 for the contribution. Looks like the examples are using the fields you made private.

AndrewReitz:
@NiteshKant How would you like to proceed? Update all the samples to use the new way or something else?

NiteshKant:
@pieces029 yep since the build fails due to the examples using the change fields, we will have to fix them before we can merge.

AndrewReitz:
@NiteshKant OK will do. So you are OK with the breaking changes?

NiteshKant:
@pieces029 generally for breaking API changes we follow a deprecation -> removal cycle spread over 2 releases but this seems to be a blocking issue for some environment and the breaking change is in constants that can be created by user, so it does not look that disruptive.

Issue #237, opened 2014-09-29
Empty Client Request Buffers Are Not Released


alexwen:
Turning on [ResourceLeakDetector#PARANOID](http://netty.io/5.0/api/io/netty/util/ResourceLeakDetector.Level.html#PARANOID); I was getting warnings from some of my tests which utilize an RxNetty HTTP Client:

```
10:45:16.876 [rx-selector-1-2] ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected.
Recent access records: 0
Created at:
    io.netty.buffer.CompositeByteBuf.<init>(CompositeByteBuf.java:59)
    io.netty.buffer.Unpooled.compositeBuffer(Unpooled.java:355)
    io.netty.handler.codec.http.HttpObjectAggregator.decode(HttpObjectAggregator.java:151)
    io.netty.handler.codec.http.HttpObjectAggregator.decode(HttpObjectAggregator.java:52)
    io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:89)
    io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333)
    io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319)
    io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:163)
    io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:148)
    io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333)
    io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319)
    io.netty.handler.logging.LoggingHandler.channelRead(LoggingHandler.java:283)
    io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333)
    io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319)
    io.netty.channel.ChannelInboundHandlerAdapter.channelRead(ChannelInboundHandlerAdapter.java:86)
    io.reactivex.netty.metrics.BytesInspector.channelRead(BytesInspector.java:56)
[...]

PR #238, opened 2014-09-29
Release empty CompositeBuffers

alexwen:
Fix for #237 

Release content if not readable.

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
@@ -127,6 +127,9 @@ public class ClientRequestResponseConverter extends ChannelDuplexHandler {
             if (content.isReadable()) {
                 invokeContentOnNext(content, stateToUse);
             }
+            else if (content.refCnt() == 1) {
+                ReferenceCountUtil.release(content);
+            }
             if (LastHttpContent.class.isAssignableFrom(recievedMsgClass)) {
                 eventsSubject.onEvent(HttpClientMetricsEvent.RESPONSE_RECEIVE_COMPLETE,
                                       Clock.onEndMillis(stateToUse.responseReceiveStartTimeMillis));


NiteshKant:
Thanks @alexwen for digging into this leak, I myself was behind this for a while but not able to point to the cause.

Looking at the code, it appears to me that the check `if (content.isReadable())` is unnecessary and we should call `invokeContentOnNext()` unconditionally. What do you think?

PS: Apologies for the delay in response, I was out on a vacation.

alexwen:
It would be my preference as well to invokeContentOnNext unconditionally. It may break backwards compatibility in cases where folks might always be expecting the content to be readable, but it is a better behavcior, imo, to have it always emit one ByteBuf.

The current behavior has forced me to put...

```
response.getContent()
                    .firstOrDefault(UnpooledByteBufAllocator.DEFAULT.buffer(0, 0))
```

... throughout my code so I can do error handling irrespective of whether or not the server has returned a message.

NiteshKant:
> > > It may break backwards compatibility in cases where folks might always be expecting the content to be readable, but it is a better behavcior, imo, to have it always emit one ByteBuf.

I agree, this is a better behavior. Do you want to update this PR with this change?

alexwen:
Attempting this change leads to other issues downstream, it looks like they are originating from ServerSentEvents (as the observable now receives the empty byte buffer before the closing of the stream). I don't have time to look at this too deeply at the moment, but will update the issue if/when I can:

```
io.reactivex.netty.protocol.http.client.HttpClientTest > testNonChunkingStream FAILED
    java.lang.ClassCastException

io.reactivex.netty.protocol.http.client.HttpClientTest > testChunkedStreaming FAILED
    java.lang.ClassCastException

io.reactivex.netty.protocol.http.client.HttpClientTest > testMultipleChunks FAILED
    java.lang.ClassCastException

io.reactivex.netty.protocol.http.client.HttpClientTest > testMultipleChunksWithTransformation FAILED
    java.lang.ClassCastException
        Caused by: rx.exceptions.OnErrorThrowable$OnNextValue

io.reactivex.netty.protocol.http.client.HttpRedirectTest > testRedirectWithConnPool FAILED
    java.util.concurrent.ExecutionException at HttpRedirectTest.java:161
        Caused by: java.util.NoSuchElementException
```

NiteshKant:
@alexwen I have created this PR #242 for the fix to send content `onNext` unconditionally. It was a little more involved than removing the if-check.

NiteshKant:
I am closing this PR in favor of #242, thanks for finding this out, anyways!

alexwen:
No problem. Glad to see there will be a fix soon!

Issue #202, opened 2014-08-11
Handling multiple HTTP based protocols on a single server.


NiteshKant:
In some offline discussions and on pull request #196 it came to the fore that serving multiple higher level protocols (SSE, Websockets, plain request-response) over HTTP on a single HTTP server is not very intuitive.

The following is the usecase:
- Lets assume we have a single HTTP server on port 8080
- URI /myapp/ws serves Web sockets traffic.
- URI /myapp/sse serves SSE traffic.
- URI /myapp/req-resp serves a simple HTTP based request-response traffic.

In order to implement a `RequestHandler` for the above routing capabilties, one would have to do something like:

``` java
 new RequestHandler<ByteBuf, ByteBuf>() {
     @Override
     public Observable<Void> handle(HttpServerRequest<ByteBuf> request,
                                    final HttpServerResponse<ByteBuf> response) {
         String uri = request.getUri();
         if (uri.startsWith("myapp/sse")) {
             ChannelPipeline pipeline =
                     response.getChannelHandlerContext().pipeline();
             // Configure pipeline for SSE
         } else {
             printRequestHeader(request);
             response.writeString("Welcome!!");
             return response.close(false);
         }
     }
 }
```

There are multiple issues here:
- The above is non-intuitive i.e. how to get a handle of netty's pipeline.
- The existing `PipelineConfigurator` implementations assume that they are only called at the creation of the pipeline and not incrementally during request processing. eg: The SSE configurator is:

``` java
        serverPipelineConfigurator.configureNewPipeline(pipeline);
        pipeline.addLast(SSE_ENCODER_HANDLER_NAME, SERVER_SENT_EVENT_ENCODER);
        pipeline.addLast(SSE_RESPONSE_HEADERS_COMPLETER, new ChannelOutboundHandlerAdapter() {
            @Override
            public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
                if (HttpServerResponse.class.isAssignableFrom(msg.getClass())) {
                    @SuppressWarnings("rawtypes")
                    HttpServerResponse rxResponse = (HttpServerResponse) msg;
                    String contentTypeHeader = rxResponse.getHeaders().get(CONTENT_TYPE);
                    if (null == contentTypeHeader) {
                        rxResponse.getHeaders().set(CONTENT_TYPE, "text/event-stream");
                    }
                }
                super.write(ctx, msg, promise);
            }
        });
```

[...]

PR #254, opened 2014-10-24
Comprehensive HTTP Router

jkschneider:
Here is a partial solution to Issue #202, at least for different HTTP methods.  There is no provision for websocket or SSE support here.  This is more or less a port of the excellent Vert.x RouteMatcher for RxNetty.  I do want to point out that I had to force queryParameters to be mutable on HttpRequest for this particular implementation.

### DIFF
--- a/rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
+++ b/rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
@@ -1,0 +1,250 @@
+package io.reactivex.netty.protocol.http.server;
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+import rx.Observable;
+
+import java.util.*;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class HttpRouter<I, O> implements RequestHandler<I, O> {
+    private final List<PatternBinding> getBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> putBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> postBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> deleteBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> optionsBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> headBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> traceBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> connectBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> patchBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private RequestHandler<I, O> noMatchHandler;
+
+    @Override
+    public Observable<Void> handle(HttpServerRequest<I> request, HttpServerResponse<O> response) {
+        String method = request.getHttpMethod().toString();
+
+        if("GET".equals(method))
+            route(request, response, getBindings);
+        else if("PUT".equals(method))
+            route(request, response, putBindings);
+        else if("POST".equals(method))
+            route(request, response, postBindings);
+        else if("DELETE".equals(method))
+            route(request, response, deleteBindings);
+        else if("OPTIONS".equals(method))
+            route(request, response, optionsBindings);
+        else if("HEAD".equals(method))
+            route(request, response, headBindings);
+        else if("TRACE".equals(method))
+            route(request, response, traceBindings);
+        else if("PATCH".equals(method))
+            route(request, response, patchBindings);
+        else if("CONNECT".equals(method))
+            route(request, response, connectBindings);
+        else
+            notFound(request, response);
+
+        if(!response.isCloseIssued()) // might have been issued by an overzealous handler
+            return response.close();
+        return Observable.empty();
+    }
+
+    public HttpRouter<I, O> get(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, getBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> put(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, putBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> post(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, postBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> delete(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, deleteBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> options(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, optionsBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> head(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, headBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> trace(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, traceBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> connect(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, connectBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> patch(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, patchBindings);
+        return this;
+    }
+
+    /**
+     * Specify a handler that will be called for all HTTP methods
+     * @param pattern The simple pattern
+     * @param handler The handler to call
+     */
+    public HttpRouter<I, O> all(String pattern, RequestHandler<I, O> handler) {
+        addPattern(pattern, handler, getBindings);
+        addPattern(pattern, handler, putBindings);
+        addPattern(pattern, handler, postBindings);
+        addPattern(pattern, handler, deleteBindings);
+        addPattern(pattern, handler, optionsBindings);
+        addPattern(pattern, handler, headBindings);
+        addPattern(pattern, handler, traceBindings);
+        addPattern(pattern, handler, connectBindings);
+        addPattern(pattern, handler, patchBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> getWithRegEx(String regex, RequestHandler<I, O> handler) {
+        addRegEx(regex, handler, getBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> putWithRegEx(String regex, RequestHandler<I, O> handler) {
+        addRegEx(regex, handler, putBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> postWithRegEx(String regex, RequestHandler<I, O> handler) {
+        addRegEx(regex, handler, postBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> deleteWithRegEx(String regex, RequestHandler<I, O> handler) {
+        addRegEx(regex, handler, deleteBindings);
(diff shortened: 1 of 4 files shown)


jkschneider:
Here is an example of using this mechanism to serve up both a rest endpoint and a bunch of static content:

``` groovy
RxNetty.createHttpServer(PORT,
    new HttpRouter()
        .get('/rest', { request, response -> /* write response... */ } as RequestHandler)
        .noMatch(RequestHandlerWithErrorMapper.from(
            new ClassPathFileRequestHandler('./static'),
            new FileErrorResponseMapper())
    )
).start()
```

Or if you want your static content to have a "context root" of **static**:

``` groovy
RxNetty.createHttpServer(PORT,
    new HttpRouter()
        .get('/rest', { request, response -> /* write response... */ } as RequestHandler)
        .getWithRegEx('/static.*',  RequestHandlerWithErrorMapper.from(
            new ClassPathFileRequestHandler('.'),
            new FileErrorResponseMapper())
    )
).start()
```

In both of these examples, all of the static content is located in a folder called **static** on the root of the classpath.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
```diff
@@ -0,0 +1,250 @@
...
+import io.netty.handler.codec.http.HttpResponseStatus;
+import rx.Observable;
+
+import java.util.*;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class HttpRouter<I, O> implements RequestHandler<I, O> {
+    private final List<PatternBinding> getBindings = new CopyOnWriteArrayList<PatternBinding>();
```
NiteshKant, line 12: Is there a reason for not having the method also as a pattern binding instead of the special handling?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
```diff
@@ -0,0 +1,250 @@
...
+    private final List<PatternBinding> connectBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private final List<PatternBinding> patchBindings = new CopyOnWriteArrayList<PatternBinding>();
+    private RequestHandler<I, O> noMatchHandler;
+
+    @Override
+    public Observable<Void> handle(HttpServerRequest<I> request, HttpServerResponse<O> response) {
+        String method = request.getHttpMethod().toString();
+
+        if("GET".equals(method))
+            route(request, response, getBindings);
```
NiteshKant, line 28: Looks like you are ignoring the result of the route (`RequestHandler.handle()`) call. If no one subscribes to the `RequestHandler.handle()` result, it will never execute.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
```diff
@@ -0,0 +1,250 @@
...
+        else if("TRACE".equals(method))
+            route(request, response, traceBindings);
+        else if("PATCH".equals(method))
+            route(request, response, patchBindings);
+        else if("CONNECT".equals(method))
+            route(request, response, connectBindings);
+        else
+            notFound(request, response);
+
+        if(!response.isCloseIssued()) // might have been issued by an overzealous handler
```
NiteshKant, line 48: This method should simply return the `RequestHandler.handle()` result.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
```diff
@@ -0,0 +1,250 @@
...
+        addPattern(pattern, handler, deleteBindings);
+        addPattern(pattern, handler, optionsBindings);
+        addPattern(pattern, handler, headBindings);
+        addPattern(pattern, handler, traceBindings);
+        addPattern(pattern, handler, connectBindings);
+        addPattern(pattern, handler, patchBindings);
+        return this;
+    }
+
+    public HttpRouter<I, O> getWithRegEx(String regex, RequestHandler<I, O> handler) {
```
NiteshKant, line 116: This model does not allow composing multiple match criterion, instead, it will require us to add a method as this for each combination.

It will instead be better to let user create a composed criterion and pass it to us. The DSL could look like:

```
router.forUri("/").withMethod(GET).withQueryParam("foo").value("bar").route(requestHandler);

router.forMethod(GET).withUri("/").withQueryParam("foo").value("bar").route(requestHandler);
```

and methods like that.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
```diff
@@ -0,0 +1,250 @@
...
+        return this;
+    }
+
+    public HttpRouter<I, O> noMatch(RequestHandler<I, O> handler) {
+        noMatchHandler = handler;
+        return this;
+    }
+
+    private void addPattern(String input, RequestHandler<I, O> handler, List<PatternBinding> bindings) {
+        // We need to search for any :<token name> tokens in the String and replace them with named capture groups
```
NiteshKant, line 180: I did not understand the need for doing this. Do we need to validate the regex?

rx-netty/src/test/java/io/reactivex/netty/protocol/http/server/HttpRouterSpec.groovy
```diff
@@ -0,0 +1,73 @@
+package io.reactivex.netty.protocol.http.server
+
+import io.netty.handler.codec.http.HttpMethod
+import spock.lang.Ignore
+import spock.lang.Specification
+
+class HttpRouterSpec extends Specification {
```
NiteshKant, line 7: I would like to keep the code consistent and not use a mix of groovy and java. Any specific reason you want to use groovy here?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    private void addRegEx(String input, RequestHandler<I, O> handler, List<PatternBinding> bindings) {
+        PatternBinding binding = new PatternBinding(Pattern.compile(input), null, handler);
+        bindings.add(binding);
+    }
+
+    private void route(HttpServerRequest<I> request, HttpServerResponse<O> response, List<PatternBinding> bindings) {
+        for (PatternBinding binding: bindings) {
+            Matcher m = binding.pattern.matcher(request.getPath()); // FIXME what is the difference between path and uri?
```
NiteshKant, line 205: Answering your question in the FIXME.

`request.getUri()` gives the entire URI where as `request.getPath()` gives only the path component of the URI.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
```diff
@@ -0,0 +1,250 @@
...
+            if (m.matches()) {
+                if (binding.paramNames != null) {
+                    // Named params
+                    for (String param: binding.paramNames)
+                        addQueryParam(request, param, m.group(param));
```
NiteshKant, line 210: `HttpServerRequest` is always immutable as the server should not change the received request. I am not sure about the intent of adding query parameter here as this is a router implementation.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/UriInfoHolder.java
```diff
@@ -53,6 +54,7 @@ public String getQueryString() {
     }
 
     public synchronized Map<String, List<String>> getQueryParameters() {
-        return decoder.parameters();
+        // decoder.parameters() sometimes returns an immutable map
+        return new HashMap<String, List<String>>(decoder.parameters());
```
NiteshKant, line 58: `HttpServerRequest` should never be mutable as it is the request received by the server.

rx-netty/src/test/java/io/reactivex/netty/protocol/http/server/HttpRouterSpec.groovy
jkschneider, line 7: Only because it was easier to write the test in Spock while you consider the whole package.  It can be rewritten in pure JUnit.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
jkschneider, line 180: We are extracting query parameters from the pattern, so you can define a route like so:

`router.get('/endpoint/:param1/:param2')`

In this case, `param1` and `param2` will be added as query parameters to the HTTP request by name.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
jkschneider, line 12: The lists could be turned into a multimap by method if you wish. I can change it if you like.  The point is that the same URI binding could exist across multiple HTTP methods.  Code complexity winds up being about the same either way.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
jkschneider, line 28: Doesn't the close call at the bottom of the method return an Observable?  I don't understand RxNetty well enough to understand the difference between the Observable returned by `handle()` and the Observable returned by `close()`.  Hopefully you can help illuminate the difference.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
jkschneider, line 116: Would you be OK if we supported both?  I far prefer the less verbose syntax for the 99% case, but I can certainly see your point about composition here as well.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
jkschneider, line 210: The intent here is to allow handlers to be able to receive these path parameters as query parameters without any other special handling.  This technique was stolen straight out of Vert.x.  Any other suggestions for how we might pass along the path parameters to a handle method that only takes `HttpServerRequest` and `HttpServerResponse`?

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
NiteshKant, line 28: True the `close()` call returns an `Observable` but thats the result of the `close()` invocation and does not indicate the result of `handle()`. I think we should not even call `close()` here, it should be up to the handler implementation (matching the uri) to call or not call close on the response.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
NiteshKant, line 180: Isn't this mixing the URI parsing and routing? I think we should separate it and provide it as a utility if someone wants to extract path fragments from a URI. Adding path fragments as query parameters does not sound correct to me.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
NiteshKant, line 116: If we have a criterion as to what qualifies as a specific method and what as the DSL, may be yes but it does not look like there is a clear answer on that. eg: is routing based on method and URI more common or method + URI regex or just method or just method and query parameter. It is a wide spectrum and the usage depends on the usecase. I think if we have a good DSL, it is far easier and intuitive to have a single way of creating these routes.

rx-netty/src/main/java/io/reactivex/netty/protocol/http/server/HttpRouter.java
NiteshKant, line 12: I was looking at it from a list of criterion + request handler tuple. Then you can apply these patterns  on a request in the order in which they are added and delegate it to the first pattern that matches. Am I missing something here?

NiteshKant:
@jkschneider I am closing this in favor of the discussion happening here: #286

PR #265, opened 2014-11-03
Make sure to release content buffer

brharrington:
When using HttpObjectAggregationConfigurator I
was getting a leak error if the response from
the server was empty:

```
Nov 02, 2014 6:01:41 PM io.netty.util.ResourceLeakDetector reportLeak
SEVERE: LEAK: ByteBuf.release() was not called before it's garbage-collected.
```

The difference seems to be the type of ByteBuf
for the content. HttpObjectAggregationConfigurator
will cause the type to be CompositeByteBuf
instead of EmptyByteBuf and when the response
is empty it isn't readable, but still needs to
be cleaned up.


NiteshKant:
Thanks @brharrington !

PR #275, opened 2014-11-12
netflix.rxjava -> io.reactivex

benjchristensen:
Updating to use new build plugins and adopt io.reactivex org


brharrington:
It looks like classes in package `com.netflix.server.context` were not renamed, was that intentional?

benjchristensen:
Nope it wasn't intentional, I wasn't looking for packages as all other packages were already the right name. 

@NiteshKant is there a reason that one random set of classes is in a different package name?

PR #293, opened 2014-12-13
Fix for stack overflow in CompositeHttpClient.shutdown().

david-at-aws:
The `CompositeHttpClient`'s constructor adds itself to the `httpClients` map as the default client. If you call `shutdown()` on a `CompositeHttpClient`, it ends up recursively calling `shutdown()` on itself, and you end up with a StackOverflowError. This change first calls `super.shutdown()` to shut down this client, then shuts down any other clients from the `httpClients` map.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/protocol/http/client/CompositeHttpClient.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/protocol/http/client/CompositeHttpClient.java
@@ -101,8 +101,12 @@ public class CompositeHttpClient<I, O> extends HttpClientImpl<I, O> {
 
     @Override
     public void shutdown() {
-        for (HttpClient<I, O> client : httpClients.values()) { // This map also contains the default client, so we don't need to shut the default explicitly.
-            client.shutdown();
+        super.shutdown();
+        for (HttpClient<I, O> client : httpClients.values()) {
+            // Constructor adds 'this' as the default client; special-case it to avoid stack overflow.
+            if (client != this) {
+                client.shutdown();
+            }
         }
     }
 


NiteshKant:
Thanks @david-at-aws 
Our CI system is not yet configured to use Java 8 and hence the build failed. I will merge this as soon as the CI system gets fixed.

david-at-aws:
Sounds good, thanks! I'm really enjoying RxJava, very cool project.

Speaking of build failures, [HttpClientTest.testConnectException2](https://github.com/ReactiveX/RxNetty/blob/0.x/rxnetty/src/test/java/io/reactivex/netty/protocol/http/client/HttpClientTest.java#L324) fails for me when run at work thanks to a firewall that proactively rejects outbound connections on port 81 (I get a ConnectException, not a ConnectTimeoutException). Not sure if there's a more reliable way to generate a ConnectTimeoutException though. :(

brharrington:
@david-at-aws does using a private ip work for you? Something like:

``` java
$ cat Test.java 
import java.net.*;

public class Test {
  public static void main(String[] args) throws Exception {
    Socket s = new Socket();
    s.connect(new InetSocketAddress("10.255.255.1", 81), 1000);
  }
}
$ java Test
Exception in thread "main" java.net.SocketTimeoutException: connect timed out
    at java.net.PlainSocketImpl.socketConnect(Native Method)
    at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:345)
    at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
    at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
    at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
    at java.net.Socket.connect(Socket.java:589)
    at Test.main(Test.java:6)
```

david-at-aws:
10.255.255.1:81 works for me at the moment (which is to say attempts to connect to it time out and fail :)). As far as I can tell, though, that's a perfectly reasonable private address that might land somewhere interesting in someone else's network and either succeed or get rejected?

brharrington:
Yeah, I don't know a perfect solution that is guaranteed to work anywhere, but most could probably find some unused private ip in their environment. It's also a little more polite in that the tests wouldn't try to connect to www.google.com.

david-at-aws:
Works for me. Want me to add that change to this pull request, or should I spin up a separate one?

NiteshKant:
@david-at-aws I am merging this PR. Sure if you have a solution in mind, go ahead and create another PR.

PR #295, opened 2014-12-14
move classes from com.netflix to io.reactivex

brharrington:
There were a handful of classes in rxnetty-contexts
that were still in a com.netflix.server.context package.
This change moves them into the io.reactivex.netty.contexts
package to make them consistent.

### DIFF
--- a/rxnetty-contexts/src/main/java/io/reactivex/netty/contexts/AbstractServerContextHandler.java
+++ b/rxnetty-contexts/src/main/java/io/reactivex/netty/contexts/AbstractServerContextHandler.java
@@ -15,7 +15,6 @@
  */
 package io.reactivex.netty.contexts;
 
-import com.netflix.server.context.BiDirectional;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
 import org.slf4j.Logger;
--- a/rxnetty-contexts/src/main/java/io/reactivex/netty/contexts/BiDirectional.java
+++ b/rxnetty-contexts/src/main/java/io/reactivex/netty/contexts/BiDirectional.java
@@ -14,10 +14,7 @@
  * limitations under the License.
  */
 
-package com.netflix.server.context;
-
-import io.reactivex.netty.contexts.ContextKeySupplier;
-import io.reactivex.netty.contexts.ContextsContainer;
+package io.reactivex.netty.contexts;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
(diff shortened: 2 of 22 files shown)


NiteshKant:
@brharrington the reason for the package names to be com.netflix is to have compatibility with legacy code inside netflix.

NiteshKant:
@benjchristensen / @quidryan looks like travis is not yet configured to use Java 8 (rx-netty examples requires java 8)

brharrington:
@NiteshKant is the package going to eventually change or will they always stay in com.netflix? If they are going to move, then what is the rough timeline? It seems like obvious backwards incompatible changes should be done sooner rather than later to prevent the internal compatibility issues from spreading.

That said, I mainly find it annoying because it isn't clear at a glance they are part of rxnetty and when browsing javadocs they are off in a strange place.

benjchristensen:
I agree we should have the same package prefix for all modules of this project. Right now that would be either `io.reactivex.netty` or `io.rxnetty` (since I have the rxnetty.io domain).

That said, we are debating names right now as we are thinking we may have outgrown the RxNetty name as this project is becoming more than just "Reactive Extensions for Netty". We are interacting with a few companies and considering a completely new IPC project where RxNetty and the new load balancer could live. Perhaps we can wait until that is figured out to do name changes?

UPDATE:

Looking at some of these though they definitely look like they should be changed.

benjchristensen:
> the reason for the package names to be com.netflix is to have compatibility with legacy code inside netflix.

@NiteshKant I don't understand this statement. What does legacy code within Netflix have to do with this library?

NiteshKant:
> > > That said, I mainly find it annoying because it isn't clear at a glance they are part of rxnetty and when browsing javadocs they are off in a strange place.

Agree it is annoying.

> > > @NiteshKant I don't understand this statement. What does legacy code within Netflix have to do with this library?

This module is a port from a public API inside netflix to open source. I am intending to use this is in the new and old stack and hence the persistence with the existing package name inside Netflix. If I change the package name, either I can use it in the existing stack or all users have to be migrated to this new package name.

benjchristensen:
Why is it part of RxNetty then? We shouldn't have code with com.netflix inside this project. 

If we need to keep the existing packages inside Netflix but want it here as well, let this code in io.reactivex be authoritative and wrapper delegation classes in the old package structure as a separate Netflix codebase that depends on RxNetty.

NiteshKant:
@benjchristensen @brharrington makes sense. Let me think over it a little more before I merge.

PR #302, opened 2014-12-22
Add bindings for spectator

dmuino:
Adds a plugin to generate [spectator](https://github.com/Netflix/spectator) metrics. Spectator has the benefit of being able to bind to [servo](https://github.com/Netflix/servo) or other metrics libraries such as [metrics](https://github.com/dropwizard/metrics).


NiteshKant:
Nice! Thanks @dmuino 

So, does it make sense to deprecate the servo plugin and use spectator plugin everywhere?

NiteshKant:
I talked with @dmuino offline; spectator bindings would deprecate the existing servo module.

Issue #303, opened 2014-12-23
HTTP client API changes for 0.5.0


NiteshKant:
This issue contains the API changes to HTTP client based on the issues #282 #280 #281 
### Http Client

All mutations to the client creates a new instance of the client.

``` java
public abstract class HttpClient<I, O> {

    /**
     * Creates a GET request for the passed URI.
     *
     * @param uri The URI for the request. The URI can be relative or absolute. If the URI is relative
     *            (missing host and port information), the target host and port are inferred from the {@link HttpClient}
     *            that created the request. If the URI is absolute, the host and port are used from the URI.
     *
     * @return New {@link HttpClientRequest}.
     */
    public abstract HttpClientRequest<I, O> createGet(String uri);

    /**
     * Creates a POST request for the passed URI.
     *
     * @param uri The URI for the request. The URI can be relative or absolute. If the URI is relative
     *            (missing host and port information), the target host and port are inferred from the {@link HttpClient}
     *            that created the request. If the URI is absolute, the host and port are used from the URI.
     *
     * @return New {@link HttpClientRequest}.
     */
    public abstract HttpClientRequest<I, O> createPost(String uri);

    /**
     * Creates a PUT request for the passed URI.
     *
     * @param uri The URI for the request. The URI can be relative or absolute. If the URI is relative
     *            (missing host and port information), the target host and port are inferred from the {@link HttpClient}
     *            that created the request. If the URI is absolute, the host and port are used from the URI.
     *
     * @return New {@link HttpClientRequest}.
     */
    public abstract HttpClientRequest<I, O> createPut(String uri);

    /**
     * Creates a DELETE request for the passed URI.
     *
     * @param uri The URI for the request. The URI can be relative or absolute. If the URI is relative
     *            (missing host and port information), the target host and port are inferred from the {@link HttpClient}
     *            that created the request. If the URI is absolute, the host and port are used from the URI.
     *
     * @return New {@link HttpClientRequest}.
     */
    public abstract HttpClientRequest<I, O> createDelete(String uri);

    /**
     * Creates a HEAD request for the passed URI.
     *
     * @param uri The URI for the request. The URI can be relative or absolute. If the URI is relative
     *            (missing host and port information), the target host and port are inferred from the {@link HttpClient}
     *            that created the request. If the URI is absolute, the host and port are used from the URI.
     *
     * @return New {@link HttpClientRequest}.
     */
    public abstract HttpClientRequest<I, O> createHead(String uri);

    /**
[...]

PR #309, opened 2014-12-29
Making ConnectionPoolBuilder immutable

wheaties:
This is in support of the goals outlined in #303. All modification methods should now return a new `ConnectionPoolBuilder` thereby avoiding some of the problems I encountered when I first tried to see what an immutable `HttpClient` would look like. Hope this the type of help you were looking for @NiteshKant

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/client/AbstractClientBuilder.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/client/AbstractClientBuilder.java
@@ -127,27 +127,27 @@ public abstract class AbstractClientBuilder<I, O, B extends AbstractClientBuilde
     }
 
     public B withMaxConnections(int maxConnections) {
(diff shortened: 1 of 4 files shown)


NiteshKant:
Thanks @wheaties . Yes I appreciate you helping out on this. I would like to do these changes in 0.5.x branch though.
I am yet to review these changes, give me some more time.

wheaties:
Should I be branching off of 0.5.x then instead of master? Also, this pull is more of a refactor step with the next step being to remove the `ConnectionPoolBuilder` entirely. I debated adding in the pool changes to the other impls of `RxClient` but held off as you indicated that is a more long term goal. I'll close this and reopen the pull request on 0.5.x branch.

NiteshKant:
> Should I be branching off of 0.5.x then instead of master?

Yep! Thanks.

PR #313, opened 2015-01-15
Only trigger the timeoutScheduler when not yet subscribed.

daschl:
This changeset makes sure that the (costly) delayed execution task
is schedule when noone is subscribed yet - since the whole purpose
of its existence is to dispose if not subscribed yet.

Note that a future improvement might be also to unsubscribe if one
subscribes before the cleanup actually runs.

**note that for some reason I couldn't get a valid unit test together (I didn't try very hard though). If you have any pointers on how to do it without changing too much let me know.**

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
@@ -270,10 +270,13 @@ public final class UnicastContentSubject<T> extends Subject<T, T> {
 
     @Override
     public void onNext(T t) {
-        ReferenceCountUtil.retain(t); // Retain so that post-buffer, the ByteBuf does not get released. Release will be done after reading from the subject.
+        // Retain so that post-buffer, the ByteBuf does not get released.
+        // Release will be done after reading from the subject.
+        ReferenceCountUtil.retain(t);
         state.bufferedSubject.onNext(t);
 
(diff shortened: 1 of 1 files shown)


rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
```diff
@@ -273,7 +273,7 @@ public void onNext(T t) {
         ReferenceCountUtil.retain(t); // Retain so that post-buffer, the ByteBuf does not get released. Release will be done after reading from the subject.
         state.bufferedSubject.onNext(t);
 
-        if (state.casTimeoutScheduled()) {// Schedule timeout once.
+        if (state.casTimeoutScheduled() && state.state == 0) {// Schedule timeout once and when not subscribed yet.
```
NiteshKant: I think it's safer to do `state.state == STATES.UNSUBSCRIBED.ordinal()`

This smells like a race (as state can be changed concurrently) but the state does not change from SUBSCRIBED/DISPOSED to UNSUBSCRIBE so we are safe.

daschl: Yup makes sense.. Do you have a hint on unit testing?

daschl:
@NiteshKant good now? :)

NiteshKant:
Yep, thanks @daschl !

Issue #311, opened 2015-01-15
UnicastContentSubject not releasing after timer


daschl:
Hi @NiteshKant @benjchristensen,

I think I found a bug in the UnicastContentSubject. I'm currently copying it over to the couchbase SDK and giving it a try and I found the following.

When you do not subscribe the timer fires, but it actually never frees the buffer - and I think here is why:

Normally, when one subscribes you chain in the releasing operator:

``` java
state.bufferedSubject.lift(new AutoReleaseByteBufOperator()).subscribe(subscriber);
```

But when the `disposeIfNotSubscribed()` fires, it is subscribed to without the operator:

``` java
    public boolean disposeIfNotSubscribed() {
        if (state.casState(State.STATES.UNSUBSCRIBED, State.STATES.DISPOSED)) {
            state.bufferedSubject.subscribe(Subscribers.empty()); // Drain all items so that ByteBuf gets released. // <-----

            if (null != state.onUnsubscribe) {
                state.onUnsubscribe.call(); // Since this is an inline/sync call, if this throws an error, it gets thrown to the caller.
            }
            return true;
        }
        return false;
    }
```

I changed the implementation also to 

``` java
state.bufferedSubject.lift(new AutoReleaseByteBufOperator()).subscribe(Subscribers.empty()); // Drain all items so that ByteBuf gets released.
```

and now it seems to work. I had to move out the operator one level since its private on the inner class, but it's not sharing any context anyway.

So

1) do you agree its an issue
2) want me to file a PR with a unit test?

I'm going with the modified version for now... at one point I wonder if we should merge back the subject into RxJava. The only problematic thing is probably how to reference the "releaseable" part, but we could come up with an interface that has a retain() and release() method which in the actual impls could be proxied through to netty buffers.

NiteshKant:
Thanks @daschl for reporting this. I missed the release during timeout in my latest changes (ugghh)
I will review your PR now.

>  at one point I wonder if we should merge back the subject into RxJava.

I think there is a value in providing a `UnicastDisposableSubject` in RxJava; which provides:
- Only a single subscriber (unicast)
- An option to dispose if not subscribed. The signal to dispose can be via an `Observable`

> we could come up with an interface that has a retain() and release() method which in the actual impls could be proxied through to netty buffers.

Yeah I think before-after `onNext` semantics is what we would need in the subject to have retain and release. Not sure what is a clean way of doing that, we can start a discussion on rxjava for this.

[...]

PR #314, opened 2015-01-15
Release ReferenceCounted when timer expires.

daschl:
This changeset subscribes when the "no subscriber" timer fires and
if the item is actually a reference counted item, completely releases
it and prevents it from leaking out of the pool.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
@@ -18,6 +18,7 @@ package io.reactivex.netty.protocol.http;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.util.ReferenceCountUtil;
+import io.netty.util.ReferenceCounted;
 import io.reactivex.netty.protocol.http.client.HttpClientResponse;
 import io.reactivex.netty.protocol.http.server.HttpServerRequest;
 import rx.Observable;
@@ -26,11 +27,9 @@ import rx.Subscriber;
 import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.internal.operators.BufferUntilSubscriber;
-import rx.observers.Subscribers;
 import rx.schedulers.Schedulers;
 import rx.subjects.Subject;
 import rx.subscriptions.Subscriptions;
-
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
@@ -138,7 +137,14 @@ public final class UnicastContentSubject<T> extends Subject<T, T> {
      */
     public boolean disposeIfNotSubscribed() {
         if (state.casState(State.STATES.UNSUBSCRIBED, State.STATES.DISPOSED)) {
-            state.bufferedSubject.subscribe(Subscribers.empty()); // Drain all items so that ByteBuf gets released.
+            state.bufferedSubject.subscribe(new Action1<T>() {
+                @Override
+                public void call(T t) {
+                    if (t instanceof ReferenceCounted) {
+                        ReferenceCountUtil.release(t, ((ReferenceCounted) t).refCnt());
+                    }
+                }
+            }); // Drain all items so that ByteBuf gets released.
 
             if (null != state.onUnsubscribe) {
                 state.onUnsubscribe.call(); // Since this is an inline/sync call, if this throws an error, it gets thrown to the caller.
--- a/rxnetty/src/test/java/io/reactivex/netty/protocol/http/UnicastContentSubjectTest.java
+++ b/rxnetty/src/test/java/io/reactivex/netty/protocol/http/UnicastContentSubjectTest.java
@@ -128,7 +128,7 @@ public class UnicastContentSubjectTest {
     }
 
     @Test
-    public void testByteBufRelease() throws Exception {
+    public void testByteBufReleaseWithNoTimeout() throws Exception {
         UnicastContentSubject<ByteBuf> subject = UnicastContentSubject.createWithoutNoSubscriptionTimeout();
         ByteBuf buffer = Unpooled.buffer();
         Assert.assertEquals("Created byte buffer not retained.", 1, buffer.refCnt());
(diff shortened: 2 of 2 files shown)


daschl:
fixes #311

rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
```diff
@@ -138,7 +137,14 @@ private UnicastContentSubject(final State<T> state, long noSubscriptionTimeout,
      */
     public boolean disposeIfNotSubscribed() {
         if (state.casState(State.STATES.UNSUBSCRIBED, State.STATES.DISPOSED)) {
-            state.bufferedSubject.subscribe(Subscribers.empty()); // Drain all items so that ByteBuf gets released.
+            state.bufferedSubject.subscribe(new Action1<T>() {
```
NiteshKant, line 140: To make it safer for future issues like this, I think it will be better to split the subject into an `Observable` and `Observer` with `Observable` always wired with auto-release. Something like:

``` java
        private final Observer<T> bufferedObserver;
        private final Observable<T> bufferedObservable;

        public State(Action0 onUnsubscribe) {
            this.onUnsubscribe = onUnsubscribe;
            final BufferUntilSubscriber<T> bufferedSubject = BufferUntilSubscriber.create();
            bufferedObservable = bufferedSubject.lift(new AutoReleaseByteBufOperator<T>());
            bufferedObserver = bufferedSubject;
        }
```

and then we do not store the `Subject` variable.

rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
```diff
@@ -138,7 +137,14 @@ private UnicastContentSubject(final State<T> state, long noSubscriptionTimeout,
...
+                @Override
+                public void call(T t) {
+                    if (t instanceof ReferenceCounted) {
```
NiteshKant, line 143: `ReferenceCountUtil.release()` does this check internally, so this is redundant.

rxnetty/src/test/java/io/reactivex/netty/protocol/http/UnicastContentSubjectTest.java
```diff
@@ -149,6 +149,18 @@ public void call(ByteBuf byteBuf) {
...
 
+    @Test
+    public void testByteBufReleaseWithTimeout() throws Exception {
+        UnicastContentSubject<ByteBuf> subject = UnicastContentSubject.create(100, TimeUnit.MILLISECONDS);
+        ByteBuf buffer = Unpooled.buffer();
+
+        subject.onNext(buffer);
+        subject.onCompleted();
+
+        Thread.sleep(500);
```
NiteshKant, line 160: This can be better done by passing a [`TestScheduler`](http://reactivex.io/RxJava/javadoc/rx/schedulers/TestScheduler.html) to `UnicastContentSubject.create()`

NiteshKant:
Thanks @daschl for the fix. Let me know what your views are on my comments.

rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
daschl, line 143: But how would you get the total refCnt out?

rxnetty/src/test/java/io/reactivex/netty/protocol/http/UnicastContentSubjectTest.java
daschl, line 160: Makes sense :)

rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
daschl, line 140: Makes sense, I'll try to modify it

daschl:
Great! Will make the changes.

Btw, we also should unsubscribe from the timer when a subscrption comes along, right?

rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
NiteshKant, line 143: I think we should release it just once as we only retain once. So, you can simply use `ReferenceCountUtil.release(Object msg)` method.

daschl, line 143: No, at this point it will have a refCnt of 2!

This is what I was wondering as well..the bytebuf is created with 1.. when it comes into this subject it is retained so it has a refcnt of 2. if noone picks it up releasing just once will not help since it will stick at 1.. I wonder if we should retain it at all?

If we don't retain it here:
- we can release() if noone subscribes
- it gets properly released if one subscribes and calls release()

Is there a specific reason it gets retained in the subject?

NiteshKant:
> Btw, we also should unsubscribe from the timer when a subscrption comes along, right?

Yes, that will remove the redundant tick from the scheduler. Makes sense.

rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
NiteshKant, line 143: > Is there a specific reason it gets retained in the subject?

Good question!
The reason is issue #264 `ByteBuf` is released after calling `onNext` on this subject. So, in case we buffer (no eager subscription to this subject), after `onNext` on this subject returns, the `ByteBuf` will have refCnt = 0 if we don't retain first.

There are two distinct scenarios:
- Eager subscription; we don't buffer and the subject is a pass through. The subject retain once, release once. The code invoking `onNext` on this Subject releases when `onNext` returns.
- No eager subscription; we buffer after doing a `retain`. If we don't retain, `ByteBuf` after `onNext` returns, ref count will be 0. After buffering, whenever this subject is subscribed to (dispose does an empty subscriber), the `ByteBuf` should be released so that user code is agnostic to this buffering (w/o buffering, user does not explicitly release)

After Issue #264 is fixed (its a behavior change), we will be removing the retain but still do release in dispose, just once.

NiteshKant:
@daschl ping, are you working on this? I would like to fix this before I do a release for 0.4.5

daschl:
@NiteshKant I can revisit it this week hopefully, but I'm a little dragged into something else. So if you want to pick it up and drag it over the finish line please go ahead :)

NiteshKant:
Yep, lemme do that. Thanks!

NiteshKant:
Closing this in favor of PR #320

Issue #280, opened 2014-11-25
Eliminate Client and Server Builders


NiteshKant:
Today, all clients and servers follow the builder pattern to make sure that a client/server once created in immutable. In order to make the creation of most common clients/server easy, `RxNetty` factory class has some factory methods, like:

``` java
    public static RxServer<ByteBuf, ByteBuf> createTcpServer(final int port,
                                                             ConnectionHandler<ByteBuf, ByteBuf> connectionHandler);

    public static HttpServer<ByteBuf, ByteBuf> createHttpServer(int port, RequestHandler<ByteBuf, ByteBuf> requestHandler);
```

It also has the following builder creation methods:

``` java
public static <I, O> ServerBuilder<I, O> newTcpServerBuilder(int port, ConnectionHandler<I, O> connectionHandler);
public static <I, O> HttpClientBuilder<I, O> newHttpClientBuilder(String host, int port);
```

The above two concepts are a bit odd as a user has to unlearn one to learn the other. eg: If one needs to enable wire-level logging, there is no way, one can do that while creating a client/server. One, has to use the builder.
### Proposal
- Fold all builder methods into the main client/server. Every invocation of those methods, should create a new instance of client/server.
- Replace inheritance with delegation in clients/servers. In absence of this change, the client/server have to introduce a generic parameter which will guide what to return from these methods which create a new client/server. The downside is that an `HttpClient` can not be used as an `RxClient`. The usage of this aspect is very limited as once, a user uses an `HttpClient`, there is very little use of using `#connect()` on the client.

After this change, users of a particular client/server would just have to be aware of any APIs on that client/server and nothing else. There will be less abstractions and less confusion.

benjchristensen:
Can you provide examples of what the new API would be and code examples of using them?

NiteshKant:
#303 details the API changes discussed here with examples.

dimitarg:
I guess the downside of this approach is that once one is done with
configuring the server, all these methods you never need to use anymore get
in the way while you're interacting with the built server. In academic
speak it violates SRP. But I'm not sure how to improve this while
addressing your original concerns.
As an end user I like it one way or the other, usually my building code
tends to be about 10 lines in one place in the app. Still I wanted to point
out.

NiteshKant:
@dimitarg Thanks for sharing your views, much appreciated!

[...]

PR #322, opened 2015-02-02
Change visibility of HttpConnectionHandler

mgajdos:
I am trying to integrate [Jersey](https://jersey.java.net) (JAX-RS RI) with RxNetty, both server and client. On the server-side I am struggling with firing some Jersey specific events (basically one event after the server container starts and the other one before the container is stopped). My approach is to extend both HttpServerBuilder and HttpServer just to override #start and #shutdown methods on HttpServer. Unfortunately this is not possible at the moment due to package-private visibility of HttpConnectionHandler (to enable RxNetty-like configuration of Jersey RxNetty container, see HttpServerBuilder#createServer()). If my approach is wrong in some way, please, let me know.

Changed visibility of:
- HttpConnectionHandler
- HttpServerBuilder.contentSubscriptionTimeoutMs

Added HttpServerExtensionTest to check that "hooks" are working in extended server.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/protocol/http/server/HttpConnectionHandler.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/protocol/http/server/HttpConnectionHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Netflix, Inc.
+ * Copyright 2015 Netflix, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,9 +26,11 @@ import rx.Observable;
 import rx.Subscriber;
 
 /**
-* @author Nitesh Kant
-*/
-class HttpConnectionHandler<I, O> implements ConnectionHandler<HttpServerRequest<I>, HttpServerResponse<O>> {
+ * Default implementation of {@link io.reactivex.netty.channel.ConnectionHandler} for HTTP protocol.
+ *
+ * @author Nitesh Kant
+ */
+public class HttpConnectionHandler<I, O> implements ConnectionHandler<HttpServerRequest<I>, HttpServerResponse<O>> {
 
     private ErrorResponseGenerator<O> responseGenerator = new DefaultErrorResponseGenerator<O>();
 
@@ -36,11 +38,11 @@ class HttpConnectionHandler<I, O> implements ConnectionHandler<HttpServerRequest
     private final boolean send10ResponseFor10Request;
     @SuppressWarnings("rawtypes")private MetricEventsSubject eventsSubject;
 
-    public HttpConnectionHandler(RequestHandler<I, O> requestHandler) {
+    protected HttpConnectionHandler(RequestHandler<I, O> requestHandler) {
         this(requestHandler, false);
     }
 
-    public HttpConnectionHandler(RequestHandler<I, O> requestHandler, boolean send10ResponseFor10Request) {
+    protected HttpConnectionHandler(RequestHandler<I, O> requestHandler, boolean send10ResponseFor10Request) {
         this.requestHandler = requestHandler;
         this.send10ResponseFor10Request = send10ResponseFor10Request;
     }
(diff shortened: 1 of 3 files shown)


mgajdos:
I am not sure but I don't think the failed verification is related to the change. Executed verification on my machine and everything passes.

NiteshKant:
@mgajdos yes the failed verification is a flaky test, it's not coz of your changes.

I did not quiet understand the relationship between overriding `start()` and `shutdown()` and making `HttpConnectionHandler` public. 

I think there is value in providing a lifecycle `Observable` for the server that someone can subscribe to receive callback of start and shutdown. I would prefer that instead of overriding start and shutdown. 

I had implemented JAX-RS extension over rxnetty for [karyon](https://github.com/Netflix/karyon) [here](https://github.com/Netflix/karyon/tree/master/karyon2-jersey-blocking) and do not remember facing issues with the implementation in regards to start() and shutdown() callbacks. Can you elaborate on your approach?

mgajdos:
There was no need to override start/shutdown methods (or provide any other mechanism that would fire such events) in Jersey 1.x. However in Jersey 2 we added new [Monitoring Feature](https://jersey.java.net/documentation/latest/monitoring_tracing.html#monitoring) and some parts (e.g. monitoring statistics) of the feature depend on startup/shutdown events being fired (via [ContainerLifecycleListener](https://jersey.java.net/apidocs/latest/jersey/org/glassfish/jersey/server/spi/ContainerLifecycleListener.html)). This is the reason I was looking for a place to fire startup/shutdown events from. I chose to override `HttpServer` methods which was the easiest way for me and I agree it's not the best solution.

The reason to make `HttpConnectionHandler` public is to give users means to configure their server as if they were using `RxNetty` class. In Jersey 2.x all container extensions use `*ServerFactory` entry point to create servers (e.g. [GrizzlyHttpServerFactory](https://github.com/jersey/jersey/blob/master/containers/grizzly2-http/src/main/java/org/glassfish/jersey/grizzly2/httpserver/GrizzlyHttpServerFactory.java#L76)). For RxNetty I wanted not only to return a server instance (`HttpServer`) but also a builder instance (`HttpServerBuilder`). For example:

```
public static HttpServerBuilder<ByteBuf, ByteBuf> createHttpServerBuilder(final int port) {
    return ...
}
```

... so users would be able to configure the server in a way they want. With this approach I'd need to call the same constructor the `HttpServerBuilder` does in its [HttpServerBuilde.createServer()](https://github.com/ReactiveX/RxNetty/blob/0.x/rxnetty/src/main/java/io/reactivex/netty/protocol/http/server/HttpServerBuilder.java#L124) method. When I am thinking about it now it'd make more sense to add a new public constructor that takes also `contentSubscriptionTimeoutMs` instead making `HttpConnectionHandler` public and make the protected constructors in `HttpServerBuilder` package private to avoid confusion.

I agree that your idea about providing an `Observable` emitting lifecycle events is better than my former approach. I'd suggest to create a new API for server that would be separated from `MetricsEvent`s (since they seems to be more request oriented). What do you think?

NiteshKant:
> However in Jersey 2 we added new Monitoring Feature and some parts (e.g. monitoring statistics) of the feature depend on startup/shutdown events being fired (via ContainerLifecycleListener). 

That's nice. Thanks I did not know that.

I am working on [milestone 0.5.0](https://github.com/ReactiveX/RxNetty/milestones/0.5.0); which for one of the awkwardness between builder and client/server as you mentioned, is intending to remove the builders (Issue #280) after which one should be able to change(create new instances) an HTTP server after creation in the same way as it is done today via the builder. I am also reasonably sure that subscription timeout isn't the correct thing to do, so I would be removing it (Issue #323).

If this is not critical I would suggest you to wait till 0.5.0 so that your efforts do not get wasted (due to underlying API changes). If you can suggest what else would be required for jersey integration, I would be happy to talk about it, feel free to open issues.

mgajdos:
The integration isn't critical, we've been asked to provide container/connector for RxNetty by some of our users, and we can wait till 0.5.0. I'll be watching the progress on 0.5.0 and if I am aware of anything that would help/affect Jersey integration I'd let you know.

I can open an issue to discuss mentioned lifecycle observable and I am also willing to participate on that task if you want.

Closing this PR as invalid.

NiteshKant:
> I can open an issue to discuss mentioned lifecycle observable and I am also willing to participate on that task if you want.

Absolutely it will be great if you can contribute. I have created [0.5.x branch](https://github.com/ReactiveX/RxNetty/tree/0.5.x), so you can send the PR to that branch.

PR #327, opened 2015-02-06
HttpServerRequest.getPath should only return path

brharrington:
If an absolute URI is passed in the request line, e.g.:

```
GET http://localhost:8090/hello HTTP/1.1
```

Then the HttpServerRequest.getPath method was returning
the full URI up to the query portion. If the user wanted
the full URI they can call HttpServerRequest.getUri. It
seems confusing to have getPath return the scheme and
host.

This change updates the UriInfoHolder used by
HttpServerRequest to have getPath only return the path.
Note, it does not use java.net.URI to do the parsing because
it is strict and fails for many URIs that are common in
practice, e.g. the URIs with `|` in the query.


NiteshKant:
Thanks @brharrington !

PR #334, opened 2015-02-18
Unsubscribe from auto release timeout if a subscriber comes along.

daschl:
If the first item gets emitted very quickly, the timer will basically
always be fired. If the timeout is very long, then all those subscriptions
will land in the old generation and very expensive to pick up.

This changeset unsubscribes immediately if a subscriber comes along, making
it more likely that the subscription is still in the young gen and quicker
to clean up.


daschl:
@NiteshKant I made this change, because it has severe impact on GC behaviour (here tested with 70s delay). Of course it gets better in general  if you reduce it to a few seconds only. 

Before:

screen shot 2015-02-18 at 11 08 48

After:

screen shot 2015-02-18 at 11 08 54

NiteshKant:
Thanks @daschl !

PR #341, opened 2015-03-18
UnicastContentSubject unsubscribes eagerly

NiteshKant:
`UnicastContentSubject` delegates buffering to a `BufferUntilSubscriber` subject. However, it subscribes the downstream subscriber to this subject using `Observable.unsubscribe()`. This causes a `SafeSubscriber` subscribing to the buffer subject which unsubscribes on termination of the upstream source.

The following code reproduces the issue:

``` java
        UnicastContentSubject<Long> subject = UnicastContentSubject.createWithoutNoSubscriptionTimeout();

        subject.onNext(1l);
        subject.onCompleted();

        subject.flatMap(aLong -> Observable.never()
                                           .doOnUnsubscribe(() -> System.out.println("unsubscribed")))
               .toBlocking().toFuture().get();
```

In the above code the flatmap returns `Observable.never()` which should not get unsubscribed, unless the eventual subscriber unsubscribes.

Using `unsafeSubscribe()` to subscribe to the `BufferUntilSubscriber` will not eagerly unsubscribe.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/protocol/http/UnicastContentSubject.java
@@ -156,11 +156,7 @@ public final class UnicastContentSubject<T> extends Subject<T, T> {
 
         private final Action0 onUnsubscribe;
 
-        public State() {
-            this(null);
-        }
-
-        public State(Action0 onUnsubscribe) {
+        private State(Action0 onUnsubscribe) {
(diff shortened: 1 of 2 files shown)


NiteshKant:
This impacts cases when the lifetime of the child `Observable` returned from `flatmap` is more than the lifecycle of the inbound request content stream.

diptanu:
LGTM. Get a release out for me :)

kurzweil:
I'm able to confirm that this fixed resolved the interrupted thread issue in my app.

PR #346, opened 2015-04-13
fix #344 read timeout not respected by client

### LINKED ISSUE #344: ReadTimeout is not respected by the client reusing a pooled connection
Here is the unit test for io.reactivex.netty.protocol.http.client.HttpClientPoolTest.java

Seems like a major issue to me - leads to the scenario when request to a dead/non-responding server could be stuck forever.

```
    @Test(expected=ReadTimeoutException.class)
    public void testReadtimeoutForReusedConnection() throws Exception {
        HttpClient.HttpClientConfig conf = new HttpClient.HttpClientConfig.Builder().readTimeout(1, TimeUnit.SECONDS).build();
        client = newHttpClient(1, PoolConfig.DEFAULT_CONFIG.getMaxIdleTimeMillis(), conf);
        //first make a normal request with no timeout to let it get to the pool
        submitAndWaitForCompletion(client, HttpClientRequest.createGet("/"), null);
        //now try the pooled connection with response delay > timeout
        submitAndWaitForCompletion(client, HttpClientRequest.createGet("test/timeout?timeout=10000"), null);
    }
```

jjteo74:
Not sure this is the proper fix but I'm using this patch for my own project for now.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/pipeline/ReadTimeoutPipelineConfigurator.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/pipeline/ReadTimeoutPipelineConfigurator.java
@@ -19,12 +19,14 @@ import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandler;
 import io.netty.channel.ChannelOutboundHandlerAdapter;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.ChannelPromise;
 import io.netty.handler.timeout.ReadTimeoutHandler;
 import io.netty.util.concurrent.EventExecutor;
 import io.reactivex.netty.protocol.http.client.ClientRequestResponseConverter;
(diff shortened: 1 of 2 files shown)


NiteshKant:
@jjteo74 Thanks for the PR. 

Looking at the read timeout implementation more, it is sub-optimal and has the following problems:
- The handler is given a `handlerAdded` callback for every write. For streaming writes, this is redundant and an overhead.
- The listener on removal still listens to channel reads and does a system call for nano time.

Seeing these issues makes me think that we should not use netty's `ReadTimeoutHandler` as is and write our own (taking most of the code from netty) and provide a way to suspend/resume based on custom events sent to the handler.
If you are interested in picking that up and the change makes sense to you, let me know, else I will start work on it.

jjteo74:
@NiteshKant Unfortunately I am quite tied up at work. Should I close this PR? (sorry, still new to Github)

NiteshKant:
@jjteo74, no worries, I will work on it. I will close this PR when I have the solution ready. Thanks!

NiteshKant:
Closing this in favor of #356 in which I have taken the code from netty's `ReadTimeoutHandler` and modified it to make it usable with connection pooling.

PR #362, opened 2015-05-29
WebSockets implementation

NiteshKant:
Websockets client/server implementation.

Also, refactored SSE client to move SSE selection on response instead of request.


NiteshKant:
The test that failed is flaky. I am merging this and will fix the test flakiness.

PR #364, opened 2015-06-01
Bug fix in Pooled connection reuse.

NiteshKant:
The reuse subscriber, in case of reuse was asynchronously sending onNext() but synchronously sending terminal events.
So, the terminal events may arrive earlier than the onNext event. Now, wrapping it over an appropriate subscriber to buffer terminal events if onNext has not yet arrived.

Also, changed ContentWriterImpl to use unsafeSubscribe instead of subscribe (duhhh)


NiteshKant:
Merging, the two tests are flaky.

PR #385, opened 2015-07-02
Fixes #376

### LINKED ISSUE #376: Provide a way to get all headers from a `HttpClientRequest`
###### Version: 0.5.x

Similar to `HttpClientResponse.headersIterator()` there should be a way to get all headers for a client request.

NiteshKant:
Providing `getHeaderNames` and `headerIterator` to `HttpClientRequest`


brharrington:
What do you think about having a common interface for HttpClientRequest, HttpClientResponse, HttpServerRequest, and HttpServerResponse? The interface could define the common aspects like methods for working with headers or the payload that would be common to any http message. I think that would help ensure consistency across the API.

NiteshKant:
@brharrington yes the read ops can be defined in a common interface, write ops returns the containing class as the operations usually create a copy of the containing class. I have horrors of multiple levels of generics in the common interface for write ops :)

PR #389, opened 2015-07-10
Added server latency and total request processing times to HttpClientEventListener

NiteshKant:
It is useful for clients to get server side latency per request which is the duration between request write completion and response recieve start.
Added this duration to responseHeaderReceived() callback.

requestProcessingComplete() was not called, fixed it now. It now gives the duration between start of header write and completion of response.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/protocol/http/client/events/SafeHttpClientEventsListener.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/protocol/http/client/events/SafeHttpClientEventsListener.java
@@ -67,9 +67,9 @@ final class SafeHttpClientEventsListener extends HttpClientEventsListener implem
     }
 
     @Override
-    public void onResponseHeadersReceived(int responseCode) {
+    public void onResponseHeadersReceived(int responseCode, long duration, TimeUnit timeUnit) {
         if (!completed.get()) {
-            delegate.onResponseHeadersReceived(responseCode);
+            delegate.onResponseHeadersReceived(responseCode, duration, TimeUnit.MILLISECONDS);
         }
     }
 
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
and I'm hoping timeunit is always millis :)

NiteshKant:
@adriancole yes it is millis as of now, although I have a PR on the way to convert them to nanos. I am assuming you were referring to the granularity _atleast_ to be millis?

codefromthecrypt:
I was half-joking, but yeah millis, but supporting micros is better (ex
zipkin trace is micros). Nanos is cool though. Even if that granularity is
not needed, you have a micro optimization: not needing to divide by 1000 to
store the result. (I am not primarily driven by this, just heard this
rationale last week)

rxnetty/src/main/java/io/reactivex/netty/protocol/http/client/events/SafeHttpClientEventsListener.java
```diff
@@ -67,9 +67,9 @@ public void onRequestWriteFailed(long duration, TimeUnit timeUnit,
     }
 
     @Override
-    public void onResponseHeadersReceived(int responseCode) {
+    public void onResponseHeadersReceived(int responseCode, long duration, TimeUnit timeUnit) {
         if (!completed.get()) {
-            delegate.onResponseHeadersReceived(responseCode);
+            delegate.onResponseHeadersReceived(responseCode, duration, TimeUnit.MILLISECONDS);
```
brharrington, line 72: Shouldn't this use the unit that was passed in?

NiteshKant, line 72: Absolutely, it should :)  looks like a copy-paste, will fix. Thanks for the review!

NiteshKant:
@adriancole cool, thanks for sharing!

PR #393, opened 2015-07-15
Moving all timeunits from millis to nanos.

NiteshKant:
Today all durations reported to Eventlisteners are milliseconds. This change will make it possible for more finer grained insights, if required.


codefromthecrypt:
:shipit:

PR #396, opened 2015-07-22
Fixes #395

### LINKED ISSUE #395: High object allocations when using `writeString` and `writeBytes` from a different thread.
`HttpServerResponse` and `ObservableConnection` provide methods like `writeString()`, `writeBytes()`, `write(R r, ContentTransformer<T, R> transformer)` and their variants with `*AndFlush()` that uses netty's `ByteBufAllocator` to allocate new `ByteBuf` if required.

RxNetty defaults to using a `PooledByteBufAllocator` that has a **thread local pool** of `ByteBuf`s. In case where this allocator is used to allocate new buffers, from a large pool of threads (eg: `Schedulers.io()` with high concurrency), this will create `ByteBuf` pools on each of those threads. This will cause a memory bloat.
##### Reproducer

(**For this test, I had turned off off-heap buffers for easier insight using mission control, the property is `-Dio.netty.noPreferDirect=true`**)

I did a small test with the following two server variants:

This writes to the response from a different thread (`Schedulers.io()` in `Observable.interval()`)

``` java
        return RxNetty.createHttpServer(8090, (request, response) -> {
            return Observable.interval(1, TimeUnit.MILLISECONDS, Schedulers.io())
                             .take(20)
                             .flatMap(aTick -> response.writeStringAndFlush("Hello!!!"));
        });
```

and this writes in the subscribing thread (eventloop)

``` java
        return RxNetty.createHttpServer(8091, (request, response) -> {
            return Observable.range(1, 20)
                             .flatMap(aTick -> response.writeStringAndFlush("Hello!!!"));
        });
```

On running the below client on both the servers:

``` java

        final CountDownLatch latch = new CountDownLatch(100);

        for (int i = 0; i < 100; i++) {
            RxNetty.<ByteBuf, ByteBuf>newHttpClientBuilder("localhost", 8091)
                   .pipelineConfigurator(new HttpClientPipelineConfigurator<>())
                   .build()
                   .submit(HttpClientRequest.createGet("/"))
                   .flatMap(response -> response.getContent()
                                                .<String>map(content -> content.toString(Charset.defaultCharset())))
                   .finallyDo(latch::countDown)
                   .forEach(System.out::println);
        }

        latch.await();
```

I found the below object allocations:
###### Write in eventloop

image

image
###### Write in `Schedulers.io()`

image

image

Looking at the above, it is evident that there are way too many allocations of `PoolThreadCache$MemoryRegionCache$Entry` which is an entry in the per-thread cache of `byte[]`, in this case.

NiteshKant:
Backporting 0.5.x for the issue of memory bloat when writing bytes/string from outside eventloop.

Added a handler that converts the `String`/`byte[]` to `ByteBuf` and allowing `String` and `byte[]` to be written directly on to the channel.

This moves the `ByteBuf` allocation from the writing thread to the eventloop.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/pipeline/PrimitiveConversionHandler.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/pipeline/PrimitiveConversionHandler.java
@@ -1,0 +1,34 @@
+package io.reactivex.netty.pipeline;
+
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.ChannelHandler.Sharable;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelOutboundHandlerAdapter;
+import io.netty.channel.ChannelPromise;
+
+@Sharable
+public class PrimitiveConversionHandler extends ChannelOutboundHandlerAdapter {
+
+    public static final PrimitiveConversionHandler INSTANCE = new PrimitiveConversionHandler();
+
+    @Override
+    public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+        Object msgToWrite = msg;
+
+        if (msg instanceof String) {
+            msgToWrite = ctx.alloc().buffer().writeBytes(((String) msg).getBytes());
+        } else if (msg instanceof byte[]) {
+            msgToWrite = ctx.alloc().buffer().writeBytes((byte[]) msg);
+        } else if (msg instanceof DelayedTransformationMessage) {
+            msgToWrite = ((DelayedTransformationMessage) msg).getTransformed(ctx.alloc());
+        }
+
+        super.write(ctx, msgToWrite, promise);
+    }
+
+    public interface DelayedTransformationMessage {
+
+        Object getTransformed(ByteBufAllocator allocator);
+
+    }
+}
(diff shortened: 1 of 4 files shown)


rxnetty/src/main/java/io/reactivex/netty/pipeline/PrimitiveConversionHandler.java
```diff
@@ -0,0 +1,34 @@
...
+@Sharable
+public class PrimitiveConversionHandler extends ChannelOutboundHandlerAdapter {
+
+    public static final PrimitiveConversionHandler INSTANCE = new PrimitiveConversionHandler();
+
+    @Override
+    public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+        Object msgToWrite = msg;
+
+        if (msg instanceof String) {
```
codefromthecrypt, line 18: curious.. should this be using ByteBufUtil?

https://github.com/netty/netty/blob/master/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L441

NiteshKant, line 18: I don't actually know the difference between `ByteBufUtil.writeUtf8()` and `writeBytes(String.getBytes())` but I assume functionally they are equivalent. (assuming the default char encoding is "UTF-8")

PR #408, opened 2015-09-02
Adding TCP Server interceptors.

NiteshKant:
Provides a new abstraction of `TcpServerInterceptorChain` that creates a chain starting with an interceptor and ending on a `ConnectionHandler`.  Any number of interceptors can be added in between. Javadoc on this class, copied below, explains the usage and design in detail:

```
What are interceptors?
Interceptors can be used to achieve use-cases that involve instrumentation related to any behavior of the connection, they can even be used to short-circuit the rest of the chain or to provide canned responses. In order to achieve such widely different use-cases, an interceptor is modelled as a simple function that takes one ConnectionHandler and returns another ConnectionHandler instance. With this low level abstraction, any use-case pertaining to connection instrumentation can be achieved.
Interceptor chain anatomy
        -------        -------        -------        -------        -------
       |       |      |       |      |       |      |       |      |       |
       |       | ---> |       | ---> |       | ---> |       | ---> |       |
       |       |      |       |      |       |      |       |      |       |
        -------        -------        -------        -------        -------
      Interceptor    Interceptor    Interceptor    Interceptor     Connection
           1              2              3              4           Handler

An interceptor chain always starts with an interceptor and ends with a ConnectionHandler and any number of other interceptors can exist between the start and end.
A chain can be created by using the various start*() methods available in this class, eg: start(TcpServerInterceptorChain.Interceptor), startWithTransform(TcpServerInterceptorChain.TransformingInterceptor), startWithReadTransform(TcpServerInterceptorChain.TransformingInterceptor), startWithWriteTransform(TcpServerInterceptorChain.TransformingInterceptor).
After starting a chain, any number of other interceptors can be added by using the various next*() methods available in this class, eg: next(TcpServerInterceptorChain.Interceptor), nextWithTransform(TcpServerInterceptorChain.TransformingInterceptor), nextWithReadTransform(TcpServerInterceptorChain.TransformingInterceptor) and nextWithWriteTransform(TcpServerInterceptorChain.TransformingInterceptor)
After adding the required interceptors, by providing a ConnectionHandler via the end(ConnectionHandler) method, the chain can be ended and the returned ConnectionHandler can be used with any TcpServer
So, a typical interaction with this class would look like:
TcpServer.newServer().start(TcpServerInterceptorChain.start(first).next(second).next(third).end(handler)) 
Simple Interceptor
For interceptors that do not change the types of objects read or written to the underlying connection, the interface TcpServerInterceptorChain.Interceptor defines the interceptor contract.
Modifying the type of data read/written to the Connection
Sometimes, it is required to change the type of objects read or written to a Connection instance handled by a TcpServer. For such cases, the interface TcpServerInterceptorChain.TransformingInterceptor defines the interceptor contract. Since, this included 4 generic arguments to the interceptor, this is not the base type for all interceptors and should be used only when the types of the Connection are actually to be changed.
Execution order
      -------        -------         -------        -------        -------        -------        -------
     |       |      |       |       |       |      |       |      |       |      |       |      |       |
     |       | ---> |       | --->  |       | ---> |       | ---> |       | ---> |       | ---> |       |
     |       |      |       |       |       |      |       |      |       |      |       |      |       |
      -------        -------         -------        -------        -------        -------        -------
       Tcp         Connection      Interceptor    Interceptor    Interceptor    Interceptor     Connection
      Server         Handler           1              2              3              4             Handler
                   (Internal)                                                                     (User)

The above diagram depicts the execution order of interceptors. The first connection handler (internal) is created by this class and as is returned by end(ConnectionHandler) method by providing a ConnectionHandler that does the actual processing of the connection. TcpServer with which this interceptor chain is used, will invoke the internal connection handler provided by this class.
The interceptors are invoked in the order that they are added to this chain.
```

Also added are two examples of how these interceptors are used.

### DIFF
--- a/rxnetty-common/src/main/java/io/reactivex/netty/channel/AbstractDelegatingConnection.java
+++ b/rxnetty-common/src/main/java/io/reactivex/netty/channel/AbstractDelegatingConnection.java
@@ -1,0 +1,339 @@
+/*
+ * Copyright 2015 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package io.reactivex.netty.channel;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.FileRegion;
+import io.netty.handler.codec.MessageToMessageEncoder;
+import io.netty.util.AttributeKey;
+import io.netty.util.concurrent.EventExecutorGroup;
+import io.netty.util.internal.TypeParameterMatcher;
+import io.reactivex.netty.HandlerNames;
+import rx.Observable;
+import rx.annotations.Beta;
+import rx.functions.Action1;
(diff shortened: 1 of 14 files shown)


NiteshKant:
Closing this PR as the interceptor changes should be on a different branch.

NiteshKant:
#409 is a replacement for this PR.

PR #421, opened 2015-09-15
Adding an Eventloop backed RxJava scheduler.

NiteshKant:
This is used to eliminate built in RxJava computation scheduler threads when running with RxNetty.
Any tasks scheduled on this scheduler for immediate execution will wakeup the underlying selector and hence will be executed immediately (as possible)

Also, provided the relevant `RxJavaSchedulersHook` to be used when setting up this scheduler as the default computation scheduler for RxJava


NiteshKant:
\cc @benjchristensen

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
```diff
@@ -0,0 +1,121 @@
...
+        this.eventLoopGroup = eventLoopGroup;
+    }
+
+    @Override
+    public Worker createWorker() {
+        final EventLoop eventLoop = eventLoopGroup.next();
+
+        return new Worker() {
+
+            private final SubscriptionList subs = new SubscriptionList();
```
akarnokd: The use of `SubscriptionList` is problematic with delayed scheduling due to O(n) removal. A more considerate container would be a regular `CompositeSubscription`. If this scheduler is dominated by non-delayed tasks, you can have both `SubscriptionList` and `CompositeSubscription`. Non-delayed tasks are then added to the list and delayed are added to the composite.

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
```diff
@@ -0,0 +1,121 @@
...
+
+            @Override
+            public Subscription schedule(final Action0 action) {
+                if (isUnsubscribed()) {
+                    return Subscriptions.unsubscribed();
+                }
+
+                /*If already on the eventloop then execute the action, else schedule it on the eventloop*/
+                if (eventLoop.inEventLoop()) {
+                    action.call();
```
akarnokd: This may be problematic if recursive scheduling happens on the eventLoop thread and may end up with stackoverflow, or at least somewhat deeper stack. You can utilize the queue in the eventLoop for breaking the recursion, or if it doesn't have one, you need to a queue-drain trampolining.

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
```diff
@@ -0,0 +1,121 @@
...
+                } else {
+                    final Future<?> result = eventLoop.submit(new Runnable() {
+                        @Override
+                        public void run() {
+                            action.call();
+                        }
+                    });
+
+                    Subscription toReturn = fromFuture(result);
+                    subs.add(toReturn);
```
akarnokd: The problem I see with this is that the task is not removed from the tracking structure after it completes which may lead to memory leak.

See http://akarnokd.blogspot.hu/2015/06/schedulers-part-2.html and perhaps http://akarnokd.blogspot.hu/2015/06/schedulers-part-3.html depending how the underlying executor is organized.

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
```diff
@@ -0,0 +1,121 @@
...
+                }
+
+                final Future<?> result = eventLoop.schedule(new Runnable() {
+                    @Override
+                    public void run() {
+                        action.call();
+                    }
+                }, delayTime, unit);
+
+                Subscription toReturn = fromFuture(result);
```
akarnokd: Same problem as in the untimed schedule.

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
NiteshKant: Thanks @akarnokd for the review. 

I did see the `EventLoopsScheduler` in `RxJava` using two subscriptions objects but wasn't clear about why use of `LinkedList` or `HashSet` would matter since I did not see an obvious random access to the subscriptions. Your distinction of delayed vs non-delayed tasks made me look at the `ScheduledAction` which is doing the `remove()`. Now in the light of that removal, using two subscription containers make sense!

This scheduler will be used instead of the regular RxJava `EventLoopsScheduler`, so predicting which type of tasks will be dominant is tough but in general I believe it will be used for timer kind of tasks as otherwise the processing happens on eventloops anyways.

Thanks for pointing this out!

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
NiteshKant: On thinking more about it, this branching actually isn't required as `eventloop.submit()` handles the `inEventLoop` case.

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
NiteshKant: So I initially "assumed" that the `ScheduledAction` is for delayed tasks but it looks like it is used for delayed and non-delayed tasks [here](https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/schedulers/NewThreadWorker.java). If so, then is the difference between them that the non-delayed tasks are executed in order (hence it is a linear removal) and delayed tasks are executed out of order and hence have random removal? And so a `LinkedList` is good for non-delayed task subscriptions?

akarnokd: > And so a LinkedList is good for non-delayed task subscriptions?

To be precise, the optimization by LinkedList relies on the observation that most non-delayed tasks come from the same thread and their add-remove order matches their submission order and thus a remove on the head of the LinkedList has less overhead than a HashMap lookup which contains more steps and may exhibit linear traversal in case of a key collision.

> delayed tasks are executed out of order and hence have random removal?

Yes, on average, a linear lookup would be worse than an amortized O(1) in a HashMap.

rxnetty-common/src/main/java/io/reactivex/netty/threads/RxJavaEventloopScheduler.java
NiteshKant: I have modified the code to follow the `EventloopScheduler` in RxJava using `ScheduledAction`

PR #422, opened 2015-09-21
Make the domain name resolver configurable

gfrison:
When a Bootstrap is created, the resolver is given by the ConnectionProvider.
The ConnectionProvider.forHost(SocketAddress) now has a second argument: NameResolverGroup
In case a proxy connection is required, create a client giving the ConnectionProvider as:
ConnectionProvider.forHost(InetSocketAddress.createUnresolved("proxied host", proxied port), NoopNameResolverGroup.INSTANCE).

### DIFF
--- a/rxnetty-common/src/main/java/io/reactivex/netty/client/ClientState.java
+++ b/rxnetty-common/src/main/java/io/reactivex/netty/client/ClientState.java
@@ -222,7 +222,8 @@ public class ClientState<W, R> extends ConnectionFactory<W, R> {
     /*Visible for testing*/ Bootstrap newBootstrap(ListenersHolder<ClientEventListener> listeners) {
         final Bootstrap nettyBootstrap = new Bootstrap().group(rawConnectionProvider.getEventLoopGroup())
                                                         .channel(rawConnectionProvider.getChannelClass())
-                                                        .option(ChannelOption.AUTO_READ, false);// by default do not read content unless asked.
+                                                        .option(ChannelOption.AUTO_READ, false)// by default do not read content unless asked.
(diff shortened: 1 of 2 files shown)


NiteshKant:
@gfrison apologies for the delayed response. The changes look good to me, although I think that providing an option to use a different resolver is better if provided on the client, what is your reason to provide it on the `ConnectionProvider`?

gfrison:
Thank you. I provided the change in `ConnectionProvider` because `ClientState` creates the `Bootstrap` instance with it.
In the Client it is possible to setup `ChannelOption` but it has different scope I suppose.
The domain resolver affects the connection, this is why `ConnectionProvider` could be a good place for it.
What do you suggest to do?

gfrison:
does make sense for you to create a builder for connection providers (`ConnectionProvider`,`PooledConnectionProvider`) in order to make them more configurable?

NiteshKant:
@gfrison the reason why I thought adding this to the client is better is because someone should not be required to alter/add a `ConnectionProvider` in order to resolve a host differently. `ConnectionProvider` uses a `ConnectionFactory` which is always provided by a client implementation, where I believe this `NameResolverGroup` will be used.

> > does make sense for you to create a builder for connection providers 

I actually do not like the way `ConnectionProvider` is designed now and I intend to change it with a simpler model soon, much on the lines of interceptors I am introducing in [this branch](https://github.com/ReactiveX/RxNetty/tree/interceptors), so would like to stay away from making changes to it, but thanks for the suggestion!

NiteshKant:
Ping @gfrison

NiteshKant:
Closing due to inactivity.

PR #423, opened 2015-09-24
upgrade to 4.1.0.Beta6

brharrington:
Updates the 0.5.x branch to use the latest netty 4.1 beta.


NiteshKant:
Thanks @brharrington. I reproduced the test failure locally and it appears that the test fails for that particular option (max messages per read) only. I am merging these changes and will fix the test in a later commit.

PR #424, opened 2015-09-24
switch to using container-based inf for travis

brharrington:
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade


brharrington:
duplicate of #403

PR #425, opened 2015-09-25
update 0.4.x to netty 4.1.0-Beta6

brharrington:
Main goal is to allow the same version of netty to be used
with both 0.4.x and 0.5.x branches.


NiteshKant:
Thanks @brharrington !

PR #428, opened 2015-09-29
fix bug removing MetricEventsListener

### LINKED ISSUE #427: (0.4.x) Can't remove a listener from a MetricEventsPubliser
# Problem description:

At one point Nitesh mentioned adding removeListener() to MetricEventsPublisher [here](https://github.com/ReactiveX/RxNetty/issues/98#issuecomment-44468564), but later changed to use Subscription instead. The rational was "returned Subscription can be unsubscribed to remove the listener". However, it seems to me thats not what actually happens.....instead calling unsubscribe only changes the state of the subscription (for example BooleanSubscription.isUnsubscribed() will return true). However, the listener (MetricsEventsPublisher) remains unchanged. 
# Additional info from Nitesh

Nitesh pointed out the source of the bug [here](https://github.com/ReactiveX/RxNetty/issues/98#issuecomment-144118414)

mikeweisskopf:
A type mismatch is preventing MetricEventsSubject.subscribe from removing a listener.

fixes #427

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/metrics/MetricEventsSubject.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/metrics/MetricEventsSubject.java
@@ -130,7 +130,12 @@ public class MetricEventsSubject<E extends MetricsEvent<?>> implements MetricEve
         Subscription subscription = Subscriptions.create(new Action0() {
             @Override
             public void call() {
-                listeners.remove(listener);
+                for (SafeListener safeListener : listeners) {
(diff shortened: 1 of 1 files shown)


NiteshKant:
Thanks @mikeweisskopf, can you add a test for this?

NiteshKant:
Ping @mikeweisskopf

mikeweisskopf:
Yes, I'll put together a test case :-)

NiteshKant:
@mikeweisskopf I am merging this change for now. Feel free to send another PR for the test :)

PR #444, opened 2015-11-07
Using numerus for percentiles in spectator plugin.

NiteshKant:
Numerus gives better percentile distributions than spectator. Using it for all latency metrics.


rxnetty-spectator-tcp/src/main/java/io/reactivex/netty/spectator/LatencyMetrics.java
```diff
@@ -0,0 +1,79 @@
...
+import com.netflix.spectator.api.ValueFunction;
+
+import java.util.concurrent.TimeUnit;
+
+import static com.netflix.numerus.NumerusProperty.Factory.*;
+
+/**
+ * A latency metric for publishing various percentiles of latency which spectator does not.
+ */
+public class LatencyMetrics {
```
codefromthecrypt, line 34: neat!

rxnetty-spectator-tcp/src/main/java/io/reactivex/netty/spectator/LatencyMetrics.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private static final NumerusProperty<Integer> latency_timeInMilliseconds = asProperty(60000);
+    private static final NumerusProperty<Integer> latency_numberOfBuckets = asProperty(12); // 12 buckets at 5000ms each
+    private static final NumerusProperty<Integer> latency_bucketDataLength = asProperty(1000);
```
codefromthecrypt, line 38: curious. does this truncate or drop, or a bin for last 1000 inputs?

NiteshKant, line 38: @adriancole are you asking, what does it do if there are more samples than the bucket data length?
If so, then it drops the old data, relevant code is [here](https://github.com/Netflix-Skunkworks/Numerus/blob/master/numerus/src/main/java/com/netflix/numerus/NumerusRollingPercentile.java#L286)

codefromthecrypt, line 38: yep answered my q! thx

PR #447, opened 2015-11-19
Handling content subscription after request/response completion.

NiteshKant:
Today, it is possible to subscribe to HTTP request/response content even after the completion of the request/response.
This would result in either the subscriber never finishing or messing up subscriptions for any subsequent request/response on that connection.

For websocket, it is valid for a content to be subscribed to after the content completes but for non-upgraded requests, it is always expected to subscribe to content after header read and before content completion.

This change adds the check that a content subscription should always arrive after header receive and before content completion.

 For WS, this restriction is relaxed.


NiteshKant:
The failed test seems to be flaky, I am merging this currently as there is a fix I need in the snapshot.

PR #465, opened 2016-01-14
Fix bug in RawRequest, add method to add headers map to avoid unnecessary copying

rdegnan:


NiteshKant:
@rdegnan haven't looked at why the tests failed but thanks for the PR. I will review & merge soon.

rxnetty-http/src/main/java/io/reactivex/netty/protocol/http/client/internal/RawRequest.java
```diff
@@ -166,7 +175,7 @@ public Observable asObservable(Connection<?, ?> connection) {
         return toReturn;
     }
 
-    private HttpRequest _copyHeaders() {
+    public HttpRequest _copyHeaders() {
```
NiteshKant: I would prefer this to be private, any reason you want to make it public?

rdegnan: No, it was really just for consistency with "public HttpRequest _copyHeaders(String uri, HttpMethod method)" -- they should probably either both be public or both be private, I'm fine either way.

NiteshKant: huh, that's correct, I didn't notice the other method is public, shouldn't be. Do you mind making that private, since you are on it :)

rxnetty-http/src/main/java/io/reactivex/netty/protocol/http/client/HttpClientRequest.java
```diff
@@ -79,9 +79,9 @@ protected HttpClientRequest(OnSubscribe<HttpClientResponse<O>> onSubscribe) {
      *
      * @param contentSource Content source for the request.
      *
-     * @return An new instance of {@link Observable} which can be subscribed to execute the request.
+     * @return An new instance of the {@link HttpClientRequest} which can be subscribed to execute the request.
      */
-    public abstract Observable<HttpClientResponse<O>> writeContent(Observable<I> contentSource);
```
NiteshKant, line 84: Apologies that I did not remember this before in our offline conversations.

The reason why `write*()` methods return `Observable<HttpClientResponse<O>>` is that the request does not support writing multiple content streams. If we return `HttpClientRequest` here, a subsequent `write*()` call will overwrite the previously written content.

This method signature avoids such unintended mistakes.

Issue #466, opened 2016-01-18
ByteBuf Memory Leak


dfjones:
In production, I found indications of memory leaks and I believe I have traced it down to the following situation. 

In a project using the RxNetty 0.4.x client through Netflix Ribbon/Hystrix, on a box with high CPU contention, it appears that memory can leak in the case where the Hystrix timeout fires before the client can deliver the ByteBuf to the subscriber. I believe this is a sort of race condition in a CPU starved environment. Imagine a race between the Hystrix timer thread and the RxNetty I/O thread, if the timer thread wins, eventually the RxNetty thread gets CPU time and produces a ByteBuf that is never consumed.

I noticed that there are safeguards in the form of timeouts in `UnicastContentSubject`, however I believe there is a bug in establishing these timeouts when a new pooled connection is created. 

I was able to reproduce this issue locally in a test that reproduces the situation described above. I've augmented the latest 0.4.x RxNetty code to add additional logging/debugging information. I'm posting what I have here, starting with the ByteBuf leak report that is attached:
[bytebuf-mem-leak-report.txt](https://github.com/ReactiveX/RxNetty/files/94916/bytebuf-mem-leak-report.txt)

The last touch of the ByteBuf comes from `onNext` in an augmented `UnicastContentSubject` shown here:

``` java
@Override
    public void onNext(final T t) {
        state.lastMem = memAddress(t);
        LOG.info("onNext called state={} buf={}:{}", state, t.toString(), state.lastMem, new Exception("trace"));
        // Retain so that post-buffer, the ByteBuf does not get released.
        // Release will be done after reading from the subject.
        ReferenceCountUtil.retain(t);
        BufferUntilSubscriber<T> bufferedSubject = (BufferUntilSubscriber<T>) state.bufferedObserver;
        touch(t, "bufferedObserver is next and bufferedObserver.hasObservers()= "
                 + bufferedSubject.hasObservers()
                 + " and UnicastContentSubject = " + this
                 + " and this.state.state = " + state.state
                 + " and state.timeoutScheduled = " + state.timeoutScheduled
                 + " and noTimeout = " + noTimeout
                 + " and state.releaseSubscription = " + state.releaseSubscription
                 + " and buf =" + t.toString() + ":" + state.lastMem
                 + " and state =" + state.toString());
        state.bufferedObserver.onNext(t);

        // Schedule timeout once and when not subscribed yet.
[...]

Issue #463, opened 2016-01-09
0.4.x WebSocketClientServerTest testMessageAggregationOnClient failure


brharrington:
I occasionally see a failure for testMessageAggregationOnClient. It also seems to happen pretty reliably on travis. It looks like the failure occurs when the frames the server is sending to the client are received at the same time as the handshake response. When it fails I see:

```
    SEVERE: [id: 0x03e41a02, /127.0.0.1:48559 => localhost/127.0.0.1:35359] RECEIVED: 153B
             +-------------------------------------------------+
             |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
    +--------+-------------------------------------------------+----------------+
    |00000000| 48 54 54 50 2f 31 2e 31 20 31 30 31 20 53 77 69 |HTTP/1.1 101 Swi|
    |00000010| 74 63 68 69 6e 67 20 50 72 6f 74 6f 63 6f 6c 73 |tching Protocols|
    |00000020| 0d 0a 75 70 67 72 61 64 65 3a 20 77 65 62 73 6f |..upgrade: webso|
    |00000030| 63 6b 65 74 0d 0a 63 6f 6e 6e 65 63 74 69 6f 6e |cket..connection|
    |00000040| 3a 20 55 70 67 72 61 64 65 0d 0a 73 65 63 2d 77 |: Upgrade..sec-w|
    |00000050| 65 62 73 6f 63 6b 65 74 2d 61 63 63 65 70 74 3a |ebsocket-accept:|
    |00000060| 20 41 67 43 4c 74 2f 4b 73 43 65 2b 4b 47 35 4e | AgCLt/KsCe+KG5N|
    |00000070| 59 62 31 32 66 61 51 50 36 65 69 6b 3d 0d 0a 0d |Yb12faQP6eik=...|
    |00000080| 0a 01 0a 30 31 32 33 34 35 36 37 38 39 80 0a 41 |...0123456789..A|
    |00000090| 42 43 44 45 46 47 48 49 4a                      |BCDEFGHIJ       |
    +--------+-------------------------------------------------+----------------+
```

When it succeeds it looks like:

```
SEVERE: [id: 0x871313e9, /127.0.0.1:63690 => localhost/127.0.0.1:63688] RECEIVED: 129B
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
[...]

PR #468, opened 2016-01-20
Potential Fix for Memory Leak Issue #466

dfjones:
If a response subscription timeout is configured, pass this timeout into the ClientRequestResponseConverter so that it can create a UnicastContentSubject with a timeout. This fix addresses issue #466

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/protocol/http/client/ClientRequestResponseConverter.java
@@ -17,6 +17,9 @@
 
 package io.reactivex.netty.protocol.http.client;
 
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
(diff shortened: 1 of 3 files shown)


dfjones:
Hmm, `WebSocketClientServerTest#testMessageAggregationOnClient` appears to be failing now. Running it locally, I see it only fails on some runs. Is this test known to be flaky or is this something that will need to be addressed in this PR?

dfjones:
I see now that this same test fails for me on the 0.4.x branch locally and in the CI here: https://travis-ci.org/ReactiveX/RxNetty/builds/97081889

brharrington:
@dfjones that test was failing before, see #463.

dfjones:
@brharrington Ok, thanks!

NiteshKant:
Thanks @dfjones for the PR! 

Yeah @brharrington mentioned that test fails when the WS frame comes together with the upgrade response in the same buffer. The failure has nothing to do with those changes, this would be fixed in the next netty release.

PR #471, opened 2016-02-02
Introduce Latency-based LoadBalancer with Power-of-Two fast selection

stevegury:
First iteration of a latency-based loadbalancer inspired by Finagle.


NiteshKant:
@stevegury thanks for doing this.

I think the tests are failing because the AbstractP2CStrategy now considers negative weight to be unusable whereas the test passes 0.0 weight.

NiteshKant:
I am merging this PR and will fix the tests. Thanks!

PR #476, opened 2016-02-08
0.5.x use travis for releases

brharrington:
I think this will work for getting the bintray publishing working via travis. This is using the rxjava project plugin instead of the normal nebula project plugin, so it is a bit different than when I have done it for other Netflix builds.

This includes Rob's changes from #475.


NiteshKant:
Thanks @brharrington

Issue #259, opened 2014-10-28
Support native netty protocol as a runtime choice.


NiteshKant:
Netty's [native transport](https://github.com/netty/netty/wiki/Native-transports) is available on linux, is more performant and has optimized memory footprint.
Since, it is a choice based on the OS, it will force people to write conditional code in the client and servers to flip the transport when available. On top of it, there would not be any standardization for an application owner to switch to native transports (when available and applicable for a client/server) if required.

RxNetty should provide a way to globally use native transport for an application via static method calls in `RxNetty` factory class. This setting should follow the following rules:
- It should be considered when a server or client has _not_ explicitly chosen a transport (via the channel and eventloop implementation). 
- If the transport implementation is applicable for that server/client. UDP is not supported by the native transport.

NiteshKant:
Release [0.3.16](https://github.com/ReactiveX/RxNetty/releases/tag/0.3.16) adds this support.

There are two ways to use this feature.

#### Independent setting

One can use native transport per client/server by using the following template:

##### Server

``` java
       RxNetty.newHttpServerBuilder(8888, new RequestHandler<Object, Object>() {
            @Override
            public Observable<Void> handle(HttpServerRequest<Object> request, HttpServerResponse<Object> response) {
            return null;
            }
       })
         .channel(EpollServerSocketChannel.class)
         .eventLoop(new EpollEventLoopGroup());
      }
```

##### Client

``` java
     RxNetty.newHttpClientBuilder("localhost", 8888)
            .channel(EpollSocketChannel.class)
            .eventloop(new EpollEventLoopGroup());
```

#### Global setting

If it is required to globally turn on/off the use of native transport, `RxNetty` provides two methods:

##### Enable

`RxNetty.useNativeTransportIfApplicable()` Enables use of native transport wherever applicable (TCP/HTTP/Websockets client and servers)

##### Disable

`RxNetty.disableNativeTransport()` Disables the use of native transport. If any client/server explicitly uses the native transport, that is not effected.

#### Additional dependency

Since, this is using the [native transport of netty](https://github.com/netty/netty/wiki/Native-transports) it requires the binary for the OS, the application is intended to be used. 
In order to get the OS specific binary for linux, one can use the following gradle dependency:

```
compile group:"io.netty", name:"netty-transport-native-epoll", version:"${netty_version}", classifier: 'linux-x86_64'
```

PR #498, opened 2016-04-22
Fixes #496 - Marks netty-transport-native-epoll dependency as optional

### LINKED ISSUE #496: netty-transport-native-epoll dependency triggers warning
Hi,

I am running into an issue where having the netty-transport-native-epoll dependency in the classpath triggers a warning from the cassandra driver. Because the os specific binary is missing, which is legitimate.
see: jhipster/generator-jhipster#3428

I found this kind of problem has already been reported [here](https://groups.google.com/a/lists.datastax.com/d/msg/java-driver-user/QgwtAB7Mhn8/Yjc6XzNXF78J)
The fix is to either add the missing os specific lib or to have the library having a dependency to  netty-transport-native-epoll marks it as [optional](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) like done in the [cassandra-java-driver](https://github.com/datastax/java-driver/blob/2.1.6/driver-core/pom.xml#L77).

Is it possible to do this with RxNetty dependencies too?

The reasoning is that in all cases, the user already have to explicitly include the native binary to use the native support with RxNetty:

`compile group:"io.netty", name:"netty-transport-native-epoll", version:"${netty_version}", classifier: 'linux-x86_64'`

For reference, the dependency to netty-transport-native-epoll was introduced by #259

raphaelbrugier:
The dependency `io.netty:netty-transport-native-epoll` is now optional and is not transitively included in the classpath of RxNetty's users

To enable Netty's native support, the user should explicitly add two dependencies:
- The one now marked as optional:
  `compile group:"io.netty", name:"netty-transport-native-epoll", version:"${netty_version}"`
- and one to get the binary specific to the OS:
  `compile group:"io.netty", name:"netty-transport-native-epoll", version:"${netty_version}", classifier: 'linux-x86_64`'

See #259 for more on how to active native support

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,10 @@
  */
 buildscript {
     repositories { jcenter() }
-    dependencies { classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:2.+' }
+    dependencies {
+        classpath 'com.netflix.nebula:gradle-rxjava-project-plugin:2.+'
+        classpath 'com.netflix.nebula:gradle-extra-configurations-plugin:3.0.3'
+    }
 }
 
 description = 'RxNetty: Reactive Extensions for Netty'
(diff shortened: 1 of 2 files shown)


NiteshKant:
Thanks @raphaelbrugier The tests are failing in this branch anyways, so it isn't your change that is causing the failures on travis.

PR #515, opened 2016-06-09
Better write backpressure

NiteshKant:
Today any writes to the channel has a `request(1)` semantic that the subscriber always ever just request 1 item from the source.
This is sub-optimal if the source that is writing isn't in-memory. For out-of-process sources, it is much better to ask for more than 1 item to get batching behavior.

This change introduces the following scheme:
- Have a max buffer size (64 now) per channel.
- This demand is equally split among all active subscribers (for more than one `write()` on the channel)
- All subscribers request up to the maximum limit allowed for them (64/n).
- The subscribers then request more when the pending items go below half the max limit (64/2*n)

### DIFF
--- a/rxnetty-common/src/main/java/io/reactivex/netty/channel/BackpressureManagingHandler.java
+++ b/rxnetty-common/src/main/java/io/reactivex/netty/channel/BackpressureManagingHandler.java
@@ -235,8 +235,7 @@ public abstract class BackpressureManagingHandler extends ChannelDuplexHandler {
         if (msg instanceof Observable) {
             @SuppressWarnings("rawtypes")
             Observable observable = (Observable) msg; /*One can write heterogeneous objects on a channel.*/
-            final WriteStreamSubscriber subscriber = new WriteStreamSubscriber(ctx, promise);
-            bytesWriteInterceptor.addSubscriber(subscriber);
+            final WriteStreamSubscriber subscriber = bytesWriteInterceptor.newSubscriber(ctx, promise);
             subscriber.subscribeTo(observable);
         } else {
             ctx.write(msg, promise);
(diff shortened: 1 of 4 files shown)


rxnetty-common/src/main/java/io/reactivex/netty/channel/BackpressureManagingHandler.java
```diff
@@ -374,33 +388,56 @@ public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exceptio
...
+         * request limit by equally distributing the demand. Minimum demand to any subscriber is 1.
+         */
+        private void recalculateMaxPerSubscriber() {
+            assert channel.eventLoop().inEventLoop();
+
+            int subCount = subscribers.size();
+            perSubscriberMaxRequest = 0 == subCount ? perSubscriberMaxRequest
+                                                    : perSubscriberMaxRequest * subCount / (subCount + 1);
+
+            perSubscriberMaxRequest = Math.max(1, perSubscriberMaxRequest);
```
stevegury, line 439: This is probably ok, but there are pathological cases where you will request more than 64.

rxnetty-common/src/main/java/io/reactivex/netty/channel/BackpressureManagingHandler.java
```diff
@@ -411,6 +448,12 @@ private void requestMoreIfWritable(Channel channel) {
 
         private final ChannelHandlerContext ctx;
         private final ChannelPromise overarchingWritePromise;
+
+        private final int initialRequest;
+        private long maxBufferSize;
+        private long pending; /*Guarded by guard*/
+        private long lowWaterMark;
+
         private final Object guard = new Object();
```
stevegury, line 457: (Unrelated to that change)
The introduction of `guard` seems unnecessary, you should synchronize on `WriteStreamSubscriber.this`.

stevegury:
👍 LGTM
I'm not yet super familiar with the codebase so take that with a grain of salt.

rxnetty-common/src/main/java/io/reactivex/netty/channel/BackpressureManagingHandler.java
NiteshKant, line 439: Yes it intentionally not being strict about exactly 64 as the demand is generated by the fact that the channel is ready to write more and 64 is more or less an artificial limit which is better than 1. Moreover, since the underlying buffer is in bytes and this limit is in terms of objects, there is less value in being precise.

rxnetty-common/src/main/java/io/reactivex/netty/channel/BackpressureManagingHandler.java
NiteshKant, line 457: @stevegury yes you are correct, I would change this in a separate PR as the change isn't related.

NiteshKant:
@stevegury thanks for the review and for the offline discussion on this :)

Issue #430, opened 2015-10-08
Provide a convenience method for HttpServerRequest to collect all content.


kurzweil:
Sometimes a developer knows, upfront, that they need all the content of a request. A convenience method (getAllContent()?), in addition to the streaming version of the getContent() method, would be helpful.

NiteshKant:
+1 this would be good to have both on `HttpServerRequest` and `HttpClientResponse`. In absence of this method, one can use the following code snippet to aggregate the content:

``` java
resp.getContent()
    .collect(Unpooled::compositeBuffer,
    (composite, buf) -> {
        int readable = buf.readableBytes();
        composite.addComponents(buf);
        composite.writerIndex(composite.writerIndex() + readable);
    })
```

jamesgorman2:
I was looking at this (basic `Observable.Transformer`) and I can make sense of `<ByteBuf, ByteBuf>` collection, but `ContentSource<T>` is generic, so I can't make sense of collection of all possible emission types from it. 

Can this be treated like `HttpClientResponse#getContentAsServerSentEvents()` which is forcing the decoding to `ContentSource<ServerSentEvent>`? It looks to me like it can't since we never store the type anywhere.

I have a couple of transformers to deal with this (one is just your code above decorated a bit, the other is byte limiter to prevent non termination). If there isn't a sensible way to make this easy it might be worth while just including them or something like them in `common`.

https://github.com/jamesgorman2/RxNetty/commit/4e723f270b48a06d9149372848a662624532164c

NiteshKant:
@jamesgorman2 you are correct that this only applies to `ByteBuf` content and I agree that since this is not generic, it is better to include it in a utility class and not the main API.

I like your approach in the code you linked here, although for bounded collection (max bytes), I would throw an error if the content exceeds the limit as opposed to doing a `takeUntil`. I think you can just have a single `Transformer` which can be used for both bounded and unbounded.

jamesgorman2:
Good call. I'll tidy up the classes/tests. I'll also fix the naming/javadoc so it's nicer to use.

NiteshKant:
Closed via #524 (Thanks @jamesgorman2 !!!)

PR #524, opened 2016-06-29
Collect sequences of ByteBufs into a single ByteBuf

jamesgorman2:
Two methods are provided - collect all and collect up to max bytes. The
latter will emit an error if over max bytes are emitted from upstream.
#430


rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
```diff
@@ -0,0 +1,153 @@
...
+            @Override
+            public Observable<ByteBuf> call(Observable<ByteBuf> upstream) {
+                return upstream
+                        .compose(new TransformerLimitByteBufTotalSize(maxBytes))
+                        .compose(INSTANCE)
+                        .flatMap(new Func1<ByteBuf, Observable<ByteBuf>>() {
+                            @Override
+                            public Observable<ByteBuf> call(ByteBuf buf) {
+                                return buf.readableBytes() > maxBytes ?
+                                        Observable.<ByteBuf>error(OnErrorThrowable.addValueAsLastCause(new TooMuchDataException(), buf)) :
```
NiteshKant: `buf` should be released here as no one will access it after this.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
```diff
@@ -0,0 +1,153 @@
...
+
+        private final int maxBytes;
+
+        public TransformerLimitByteBufTotalSize(int maxBytes) {
+            this.maxBytes = maxBytes;
+        }
+
+        @Override
+        public Observable<ByteBuf> call(Observable<ByteBuf> upstream) {
+            return upstream.scan(
```
NiteshKant: If I understand this correctly, the `scan+skip+takeUntil` dance is required just to make sure we don't cross the buffer limit?
If so, then may be it is ok to just throw an `Exception` which will anyways be sent as an `onError` to the downstream, does that make sense? It will simplify the code.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
```diff
@@ -0,0 +1,153 @@
...
+ * ByteBuf.
+ */
+public class CollectBytes implements Transformer<ByteBuf, ByteBuf> {
+
+    private static final CollectBytes INSTANCE = new CollectBytes();
+
+    /**
+     * Collect all emitted ByteBufs into a single ByteBuf.
+     */
+    public static Transformer<ByteBuf, ByteBuf> all() {
```
NiteshKant: Thinking about this more, I think we can cap `all()` on `Integer.MAX_VALUE`. This will have sane limits (2GB) on the buffer and reuse the code for both methods, WDYT?

jamesgorman2: I'm neither here nor there, but this did make me realise there was an overflow bug calling `byteBuf.writerIndex(collector.byteBuf.writerIndex() + i)` if more that `Integer.MAX_VALUE` total bytes are received by `#upTo()` anyway so I rejigged the whole thing to be safer. 

It's a bit more complex inside but should be correct w.r.t. sizes.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
NiteshKant: @jamesgorman2 what's your take on this comment i.e. just do `collect()` instead of `scan+skip+takeUntil` and throw when the size limit is breached.

jamesgorman2: Sorry, notifications got buried.

I originally planned to just drop the data and throw, but I realised I sometimes want to know what I got before maxBytes was received (especially for logging - ie why am I getting too many bytes?).

This is making the error either a simple 'do not continue' or a signal that there was an overflow.

BUT, from your comment above it looks like there is a risk of frequent memory leak if users don't care about the content of the error or call `ContentSource#autoRelease()`.

On the basis of this, I think the only sensible path is roll back to a simpler fail-only model (ie what you are suggesting above).

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
jamesgorman2: see comment below on this

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
NiteshKant: Yes but you still need to `release` before throwing as nobody else will be able to access the last buffer and composite after this error.

jamesgorman2: This buf (the composite) was available from
`((OnErrorThrowable.OnNextValue) tooMuchDataException.getCause()).getValue()`
_if_ the user decides to actually get it. A diligent user could make sure they always check and release this, but it's a pretty massive hole since it's on the error path.

(there was also another buf that was not getting released that was the cause of the overflow - I think we are talking at cross purposes here through :-) )

I'm quickly putting together a proposal to close both these holes that follows your suggestion (and should close leaks!)

jamesgorman2: New version up should address these. The interface is not as nice since I can't get the bytes collected so far, but it is safer.

Also added a comment about (not) using autorelease since the whole thing dies if you do.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    public static class TooMuchDataException extends RuntimeException {
+        public TooMuchDataException() {
+            super();
+        }
+    }
+
+    private static class CountingCollector {
+        private long count;
```
NiteshKant: I think `CompositeBuffer.readableBytes()` would suffice instead of keeping this count, which in turn removes the need of this class.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
```diff
@@ -0,0 +1,125 @@
...
+                    }
+                },
+                new Action2<CountingCollector, ByteBuf>() {
+                    @Override
+                    public void call(CountingCollector collector, ByteBuf buf) {
+                        collector.count += buf.readableBytes();
+                        if (collector.count <= maxBytes) {
+                            int i = buf.readableBytes();
+                            collector.byteBuf.addComponent(buf);
+                            collector.byteBuf.writerIndex(collector.byteBuf.writerIndex() + i);
```
NiteshKant: I think now `CompositeByteBuf` has an overload for `addComponent` which takes a `increaseWriterIndex` flag, which you can use.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
```diff
@@ -0,0 +1,125 @@
...
+                        } else {
+                            collector.byteBuf.release();
+                            buf.release();
+                            throw new TooMuchDataException();
```
NiteshKant: Should we pass a message here specifying what was the limit and how much was received?

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
NiteshKant: > > Also added a comment about (not) using autorelease since the whole thing dies if you do.

Hmm ... that is unfortunate 😞 but I don't think there is a way to avoid this easily.

jamesgorman2: I did some tests with the `AutoreleaseOperator` and it looked pretty much like the only way out would be to copy bytes so I left. 

It did give me some good thoughts on safe memory handling for an `Operator<ByteBuf, Char>`, so I'll look at that as a small win

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
jamesgorman2: Good call, I'll simplify this.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
jamesgorman2: Ditto for this one.

rxnetty-common/src/main/java/io/reactivex/netty/util/CollectBytes.java
jamesgorman2: I wasn't sure about this - happy to put one in. Only makes sense to refer to maxBytes since we never end the stream.

jamesgorman2:
I've done those updates (and improved the javadoc further). Are you running auto-squash merges or should I squash this down when you're happy with it?

as an aside, it's a pity that memory but management meant that some of the nice stuff I wanted couldn't work, but it is nice to remove code :-)

NiteshKant:
@jamesgorman2 yep I do squash and merge so you don't need to squash!

> > it's a pity that memory but management meant that some of the nice stuff I wanted couldn't work, but it is nice to remove code :-)

Tradeoffs!!

Issue #523, opened 2016-06-27
Possible Connection Leak and wrong metrics


soemeier:
We have experienced some Connection leak problems when using Ribbon together with rxNetty. After some analysis we found the root cause. The load balancer took out one of the servers which lead to the connection pool to be closed (`ConnectionPoolImpl.shutdown()`).

Because the connection pool got closed the isDone Flag in the `SafeListener` was set to true. When the connection that were still in use were put back to the pool the metrics were not counted anymore. 

Therefore the liveConnections counter was not reduced but which was even worse the limitDeterminationStrategy did not not reduce the number of used connections. At some point we got PoolExaustedException. 

For me the question is how to fix this problem. I did not fully understand why the metrics are not allowed to be submitted anymore once the pool is shut down. So one option for me would be to always submit the metrics. But there is probably a reason for that.

Another option for me would be to call the `onCompleted()` on the `metricsEventsSubject` only if all connections have been returned.

I would be happy to provide a fix but I am not sure which direction is the best way to go.

NiteshKant:
> > Another option for me would be to call the onCompleted() on the metricsEventsSubject only if all connections have been returned.

@soemeier I think this would be the correct way. The metrics listener are sent an `onComplete()` as some listeners may hold state and `onComplete()` is a way to cleanup the state. I will be happy to look at your proposal to fix this. If you need help, let me know!

> > Therefore the liveConnections counter was not reduced but which was even worse the limitDeterminationStrategy did not not reduce the number of used connections. At some point we got PoolExaustedException.

Ideally `acquire()` should fail if the pool is shutdown, but I see a case where if the `Observable` returned by `acquire()` is subscribed post `shutdown`, it will not reject. May be we should fix that too.
Otherwise, these things look contradictory, if the pool is shutdown, you must not get `PoolExhaustedException` as one should never acquire post shutdown.

NiteshKant:
Fixed via #525

PR #525, opened 2016-06-30
Allow to return connections even if pool was already closed

soemeier:
Fix for #523. I think I have also fixed the problem the you mentioned if the subscriber is being called when the pool is shutDown.

### DIFF
--- a/rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
+++ b/rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
@@ -20,6 +20,10 @@ import io.reactivex.netty.channel.ObservableConnection;
 import io.reactivex.netty.metrics.Clock;
 import io.reactivex.netty.metrics.MetricEventsListener;
 import io.reactivex.netty.metrics.MetricEventsSubject;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Observable;
@@ -54,7 +58,13 @@ public class ConnectionPoolImpl<I, O> implements ConnectionPool<I, O> {
     private final RxClient.ServerInfo serverInfo;
     private final PoolConfig poolConfig;
     private final ScheduledExecutorService cleanupScheduler;
-    private final AtomicBoolean isShutdown = new AtomicBoolean();
+    private final AtomicInteger aquiredConnectionsCounter = new AtomicInteger();
+    //is needed to not accept any new connections that are aquired
+    private final AtomicBoolean isShutdownRequested = new AtomicBoolean();
+    //is needed to not perform the shutdown twice
+    private final AtomicBoolean isShutdownPerformed = new AtomicBoolean();
+    //is needed to not perform shutdown during the aquire of a connection
+    private final ReadWriteLock shutdownLock = new ReentrantReadWriteLock();
     /*Nullable*/ private final ScheduledFuture<?> idleConnCleanupScheduleFuture;
 
     /**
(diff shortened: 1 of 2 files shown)


soemeier:
Hmm I have checked those tests. They do not even do anything with the connection pool. Might it be that they are unstable?

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -110,54 +120,73 @@ public ConnectionPoolImpl(RxClient.ServerInfo serverInfo, PoolConfig poolConfig,
     @Override
     public Observable<ObservableConnection<I, O>> acquire() {
 
-        if (isShutdown.get()) {
+        if (isShutdownRequested.get()) {
```
NiteshKant: This should be moved inside the `call` method below, so that it gets checked for every subscription of the returned `Observable`.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -54,7 +58,13 @@
...
     private final PoolConfig poolConfig;
     private final ScheduledExecutorService cleanupScheduler;
-    private final AtomicBoolean isShutdown = new AtomicBoolean();
+    private final AtomicInteger aquiredConnectionsCounter = new AtomicInteger();
+    //is needed to not accept any new connections that are aquired
+    private final AtomicBoolean isShutdownRequested = new AtomicBoolean();
+    //is needed to not perform the shutdown twice
+    private final AtomicBoolean isShutdownPerformed = new AtomicBoolean();
+    //is needed to not perform shutdown during the aquire of a connection
+    private final ReadWriteLock shutdownLock = new ReentrantReadWriteLock();
```
NiteshKant, line 69: So, I don't think we need to be absolutely sure that shutdown does not happen during acquire. There are the following reasons for it:
- There could always be already checked out connections which aren't too different from concurrent acquire and shutdown.
- This lock means that the `shutdown` can not proceed while acquires are happening. If the shutdown happens from the eventloop (it can for rejection of hosts), it blocks the eventloop.

I would suggest having a boolean guarded by `synchronized` which is the flag for shutdown. If `acquire` finds this flag to be `true` before acquire, it is rejected.

`release` should cleanup when `aquiredConnectionsCounter` is 0 and the pool is shutdown.

The above approach will have some connections, when `release` happens after `shutdown` but between `acquire` checking for `shutdown` and actually acquiring, that will be checkout even when the pool is cleaned up. I think we are ok with such behaviors if we check for `shutdown` within `onSubscribe` (which I have suggested below). WDYT?

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -169,23 +198,27 @@ public void call(final Subscriber<? super ObservableConnection<I, O>> subscriber
...
 
         try {
             connection.getChannel().pipeline().fireUserEventTriggered(new PooledConnectionReleasedEvent(connection));
             metricEventsSubject.onEvent(ClientMetricsEvent.POOL_RELEASE_START);
-            if (isShutdown.get() || !connection.isUsable()) {
+            if (isShutdownRequested.get() || !connection.isUsable()) {
                 discardConnection(connection);
                 metricEventsSubject.onEvent(ClientMetricsEvent.POOL_RELEASE_SUCCESS, Clock.onEndMillis(
                         startTimeMillis));
+                performShutdownIfRequested();
```
NiteshKant: I think it will be better to call this when `aquiredConnectionsCounter.decrementAndGet();` above returns `0`

NiteshKant:
@soemeier yes the tests are flaky. Thanks for the changes

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
soemeier: It is already in there but I wanted to fail fast. `if (!lockRetrieved || isShutdownRequested.get())` Do you think it is enough in the call method?

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
soemeier, line 69: > > There could always be already checked out connections which aren't too different from concurrent acquire and shutdown.

Shouldn't it be that in that case the aquiredConnectionsCounter not equal to zero and the shutdown would not take place then?

> This lock means that the shutdown can not proceed while acquires are happening. If the shutdown happens from the eventloop (it can for rejection of hosts), it blocks the eventloop.

I see your point. Blocking in the event loop is not a good idea.

> release should cleanup when aquiredConnectionsCounter is 0 and the pool is shutdown.

That is what it is doing in my implementation or did I miss something?

> The above approach will have some connections, when release happens after shutdown but between acquire checking for shutdown and actually acquiring, that will be checkout even when the pool is cleaned up. I think we are ok with such behaviors if we check for shutdown within onSubscribe

I somehow do not like the idea that there is still the possibility of a connection leak. I know that this will be a rare case but it can still happen. What about the following approach:
Instead of blocking in the `shutdown` method we just do nothing if we can't acquire the lock. If we can't acquire the lock it basically means that someone is just about to acquire a connection. That connection will at some time be returned and the connection pool is closed.
What could still happen is that acquiring the connection fails. In that case the connection would not be returned and the pool would not be closed. So if the acquire fails we check if we have to shut down the pool. This way there is no chance to return a connection from an already closed connection pool.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
soemeier: I will change that.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
NiteshKant: Yes I think doing it on subscription is enough (and correct) as subscriptions are lazy. So, generally state checks are better on subscription vis-a-vis on creation.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
NiteshKant, line 69: I kind of like your approach if it was without the lock :) I somehow don't like the idea of locking for a non-critical operation. Having said that, it may be clearer to me if I see the code.

So, there isn't a connection leak but a discrepancy in metrics, rite?

soemeier:
I did a lot of thinking about it and came to the conclusion that the shutdown on demand just blows up the code. For readability I think it is better to create a separate thread that tries to shutdown the pool on a regular base. Not very nice since we have to create another thread but I think it is more readable. Maybe you have a better approach then using a new thread.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
soemeier, line 69: Well it is not a real leak. So the connections get returned. But ribbon uses a global counter. So all the pools share the same counter in order to have a general connection limit. If that counter does not get decremented due to wrong metrics we get pool exhausted at some point.

NiteshKant:
@soemeier apologies I haven't been able to look at your recent changes. Perhaps, will do so in a couple of days. Thanks for putting this effort to fix ❤️

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -210,17 +250,64 @@ public void call(final Subscriber<? super ObservableConnection<I, O>> subscriber
 
     @Override
     public void shutdown() {
-        if (!isShutdown.compareAndSet(false, true)) {
+        if (!isShutdownRequested.compareAndSet(false, true)) {
             return;
         }
+        boolean shutdownPerformed = performShutdownIfPossible();
+        if (!shutdownPerformed) {
+            ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor(new ThreadFactory() {
```
NiteshKant: I would rather use `Observable.timer()` instead of creating our own Schedulor (looks like for each shutdown?)

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -110,75 +123,100 @@ public ConnectionPoolImpl(RxClient.ServerInfo serverInfo, PoolConfig poolConfig,
...
+                        subscriber.onError(new IllegalStateException("Connection pool is already shutdown."));
+                        return;
                     }
-                } catch (Throwable throwable) {
-                    metricEventsSubject.onEvent(ClientMetricsEvent.POOL_ACQUIRE_FAILED,
-                                                Clock.onEndMillis(startTimeMillis), throwable);
-                    subscriber.onError(throwable);
+                    performAquire(subscriber);
+                }   finally {
+                    lock.unlock();
```
NiteshKant, line 143: Since connect is asynchronous, this `try-finally` isn't really guaranteeing that there won't be a pending `connect` when `lock` is released.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -110,75 +123,100 @@ public ConnectionPoolImpl(RxClient.ServerInfo serverInfo, PoolConfig poolConfig,
...
                 }
             }
         });
     }
 
+    private void performAquire(Subscriber<? super ObservableConnection<I, O>> subscriber) {
+        logger.trace("Started to aquire connection");
```
NiteshKant: I don't think we need these trace logs.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
soemeier, line 143: That is true but I increment the connections counter before I release the lock (in the previous commit it was a different counter but that is actually not needed). So if the lock is released the shutdown can not take place because the counter got incremented.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
soemeier: I have removed them. They just helped me to understand the problem better ;).

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
soemeier: I changed this to Observable. Seems to be the better solution. Please have a look at it.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -210,17 +243,64 @@ public void call(final Subscriber<? super ObservableConnection<I, O>> subscriber
...
     @Override
     public void shutdown() {
-        if (!isShutdown.compareAndSet(false, true)) {
+        if (!isShutdownRequested.compareAndSet(false, true)) {
             return;
         }
+        Observable.just(1L).subscribe(createShutdownAction());
+    }
+
+    private void performShutdownIfRequested() {
```
NiteshKant, line 252: Looks like this method is not used anymore.

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -273,12 +354,22 @@ public void call(ObservableConnection<I, O> connection) {
...
                                           metricEventsSubject.onEvent(ClientMetricsEvent.POOL_ACQUIRE_FAILED,
                                                                       Clock.onEndMillis(startTime), throwable);
+                                          aquiredConnectionsCounter.decrementAndGet();
                                           subscriber.onError(throwable);
                                       }
                                   }
         );
     }
 
+    private void poolAlreadyClosed(PooledConnection<I, O> connection, long startTime, Subscriber<? super ObservableConnection<I, O>> subscriber) {
```
NiteshKant, line 364: Method not used anymore?

rxnetty/src/main/java/io/reactivex/netty/client/ConnectionPoolImpl.java
```diff
@@ -289,18 +380,47 @@ public Subscription subscribe(MetricEventsListener<? extends ClientMetricsEvent<
...
+                        lock.unlock();
                     }
                 }
             } catch (Exception e) {
                 logger.error("Exception in the idle connection cleanup task. This does NOT stop the next schedule of the task. ", e);
             }
         }
     }
+
+    private class ShutdownTask implements Runnable {
```
NiteshKant, line 406: class not used anymore?

NiteshKant:
Apologies again @soemeier for the delayed responses, I am still on PTO. 
I think we can clean up the code a bit (unused classes/methods) then I can pull it in. I am putting this up for the next milestone.

NiteshKant:
@soemeier I am pulling this in and will do the cleanup myself. Thanks for the contribution and apologies for the delay due to me being on PTO.

PR #537, opened 2016-08-05
update to netty 4.1.4.Final

brharrington:
There is a regression in netty 4.1.3.Final, see:

http://netty.io/news/2016/07/27/4-0-40-Final-4-1-4-Final.html

The 4.1.4.Final version also includes a change to
QueryStringDecoder that broke some of the UriInfoHolder
tests. In particular the path is now decoded. I opted
to change the tests to match the netty behavior. For
context the netty ticket is: https://github.com/netty/netty/issues/5590

Tests were successful locally after those changes.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,5 +13,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-netty_version=4.1.3.Final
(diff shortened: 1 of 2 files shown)


NiteshKant:
Thanks @brharrington !

Issue #548, opened 2016-09-02
RxNetty SSL configuration


groodt:
It seems that for RxNetty, I have to manually specify if the client should use SSL. Something like the snippet below for 0.4.x

```
        CompositeHttpClientBuilder<ByteBuf, ByteBuf> clientBuilder = new CompositeHttpClientBuilder<>();

        if (!env.contains("local")) {
            clientBuilder.withSslEngineFactory(new DefaultSSLEngineFactory()); // SSL only enabled for non-local
        }

        return clientBuilder.build();
```

I also had to create a class DefaultSSLEngineFactory like this:

```
    private static class DefaultSSLEngineFactory implements SSLEngineFactory {

        private final SslContext sslCtx;

        private DefaultSSLEngineFactory() {
            try {
                sslCtx = SslContextBuilder.forClient().build();
            } catch (SSLException e) {
                throw new IllegalStateException("Failed to create default SSL context", e);
            }
        }

        @Override
        public SSLEngine createSSLEngine(ByteBufAllocator allocator) {
            return sslCtx.newEngine(allocator);
        }
    }
```

For ease of use, this seems a little cumbersome in my opinion. Is there a specific reason for it? Coming from using HC Apache or AsyncHttpClient/async-http-client, this seems like a lot of extra work.

I imagine for many client use-cases, they will want to communicate over SSL/TLS in Production environments, and use non-SSL/TLS for localhost, CI and/or integration test environments.

It seems that if I enable SSL as above for localhost, then communication over http:// doesn't work, so I have to explicitly unwire it for the clients on localhost.

I'm asking if I am doing something wrong, or if there is a strong rationale for this?

NiteshKant:
> > > For ease of use, this seems a little cumbersome in my opinion. Is there a specific reason for it? 

0.5.x makes the API "better" by introducing `.secure()` and [`.unsafeSecure()`](https://github.com/ReactiveX/RxNetty/blob/0.5.x/rxnetty-examples/src/main/java/io/reactivex/netty/examples/http/secure/SecureHelloWorldClient.java#L78) methods on the client. For quick testing `unsafeSecure()` is pretty handy as it uses a `trust-all` factory. 
For actual production code, it requires netty's `SSLCodec` or `SSLEngine` which I think is the norm (?).

> > > I imagine for many client use-cases, they will want to communicate over SSL/TLS in Production environments, and use non-SSL/TLS for localhost, CI and/or integration test environments.

[...]

PR #553, opened 2016-09-08
Simple example that connects to github securely.

groodt:
A simple example after discussions from here: https://github.com/ReactiveX/RxNetty/issues/548

### DIFF
--- a/rxnetty-examples/HTTP.md
+++ b/rxnetty-examples/HTTP.md
@@ -12,6 +12,7 @@ and an [Hello Server](src/main/java/io/reactivex/netty/examples/http/helloworld/
 This example constitutes of an [Hello Client](src/main/java/io/reactivex/netty/examples/http/secure/SecureHelloWorldClient.java)
 and an [Hello Server](src/main/java/io/reactivex/netty/examples/http/secure/SecureHelloWorldServer.java)
 
+- __Secure Client Default SSLContext__: A very simple example that demonstrates one way of connecting a client securely to a public HTTPS server using the system defaults [SecureDefaultHttpClient](src/main/java/io/reactivex/netty/examples/http/secure/SecureDefaultHttpClient.java)
 
 - __Simple Interception__: A example that demonstrates how to write basic interceptors for HTTP server and client.
 This example constitutes of an [Intercepting Client](src/main/java/io/reactivex/netty/examples/http/interceptors/simple/InterceptingClient.java)
--- a/rxnetty-examples/src/main/java/io/reactivex/netty/examples/http/secure/SecureDefaultHttpClient.java
+++ b/rxnetty-examples/src/main/java/io/reactivex/netty/examples/http/secure/SecureDefaultHttpClient.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2015 Netflix, Inc.
+ *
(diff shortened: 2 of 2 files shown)


NiteshKant:
Thanks @groodt !

I think it is not a good idea to refer to external sites from a test (even if it is netflix 😄 ). 
I would prefer following the current model of having a server and a client for a test. I am also OK if you update the current secure example that uses `unsafeSecure()` and use your code instead.

groodt:
@NiteshKant I do like the current way you have it setup with a server and a client, but I'm not sure there is a clean way to do this without talking to a public server.

The problem is as follows:
Some users of RxNetty might like to make secure API requests to public APIs from providers like GitHub, Google, Twitter and many others. These services are secured with certificates signed by the major certificate authorities and these public certificates get bundled with operating systems, browsers, JDKs etc. So when you use a browser or a tool like Postman or really anything, it uses these public certificates that you already have bundled. It is a very common use-case. Even _within_ an organisation you might find certificates that are not-self signed for whatever reason.

So to have a test or example that demonstrates how to use secure SSL for this common use case, I'm not sure I can actually do this for an example against localhost.

It would probably be possible to create a server with a self-signed certificate on localhost and bundle this up in the examples project, but the setup of the SSLEngine for this use-case will look different to the use-case above. 

Which gets us back into the same situation we started. That there isn't an example that demonstrates how to use RxNetty against an SSL secured API in the wild.

groodt:
Any other ideas for this? I could point it to www.howsmyssl.com instead?

It might be that this example can't live in the project. I would understand it and perhaps I'd just create my own simple project that demonstrates one way to use RxNetty client against a server in the wild. Or a Stackoverflow question.

On the other hand, RxNetty is in the business of providing an HTTP Client, so perhaps an integration test against a server outside the RxNetty ecosystem is valuable so that you can confirm that RxNetty can work against non-RxNetty servers.

jamesgorman2:
From the [peanut gallery](https://en.wikipedia.org/wiki/Peanut_gallery), a working example against a real CA certed site is good. And this give a good intro to how SSLEngine fits in. I'd be more worried about externalities if it was a real test, but setting up a dummy CA, as well as the dummy server strikes me as overkill. Thoughts on sites (in order):
- https://github.com - the source is hosted here, the issues link goes here and it's well known
- https://reactivex.io/ - since this is an RX project if you  want to show a cert failure :trollface: (looks like a signing issue with using github hosting, I have put an issue in)
- https://netty.io/ - since this is a netty adapter
- https://www.netflix.com - project sponser

NiteshKant:
Thanks @jamesgorman2 for the input.

@groodt I understand the situation and I think this is a good addition to the examples, the only thing that it would miss is a unittest as I don't feel comfortable running tests connecting to an external site, it makes the test flaky.

Can you change the target host to github? I will pull this in after that.

groodt:
Thanks for the input @jamesgorman2 and @NiteshKant 

I've updated the example to make a request against github.com

I agree that making a real unit test that tests against a real server could lead to flaky tests and random failures. No easy way around it I think. This is the external boundary of the system where the client interacts with the 'dirty' real-world.

Let's leave it as a simple example then.

NiteshKant:
Thanks @groodt for the contribution!

PR #557, opened 2016-09-12
Make SSE content type check tolerant of charset

### LINKED ISSUE #552: Check for SSE stream does not take charset into account
The check for SSE content type in `HttpClientResponseImpl` does not take into account that a charset may also be present in the header.  A content type of `text/event-stream;charset=UTF-8` results in an `IllegalStateException`.

https://github.com/ReactiveX/RxNetty/blob/1d6eb41994b70098fe0c2ec68b5ce1e650951e69/rxnetty-http/src/main/java/io/reactivex/netty/protocol/http/client/internal/HttpClientResponseImpl.java#L258

rhart:
Fixes #552


rhart:
There wasn't an obvious place to add a test for this.  AFAICT there aren't any existing tests for unexpected content types and it's not overridable when using the http client as the sse encoder overrides whatever you set.

NiteshKant:
Thanks @rhart!
Yeah adding a test will need more changes to `HttpServerResponse`.

PR #567, opened 2016-11-10
Escape html special characters in default error response generator.

tokuhirom:
RxNetty should escape HTML escape characters in default error response generator.
Current implementation may cause XSS( https://www.owasp.org/index.php/Cross-site_Scripting_(XSS) ).

If a user put a request path, query string, or any other string from client's request, it cause security vulnerability.


NiteshKant (approved):
Thanks @tokuhirom for the change!

PR #579, opened 2017-02-06
Refactor spectator metrics

NiteshKant:
__Problem__

Currently the listener does not expose metrics to specific events but instead does internal calculations and expose custom gauges. This is limiting as it gives less information.

__Modification__

Modified the metrics to represent the actual events.

__Result__

More flexible insights.


codefromthecrypt:
a lot simpler, too

PR #603, opened 2018-03-09
Fix headers in markdown that now require an intervening space

davidmoten:
github markdown changed a few months ago and now requires an intervening space after ### headers


jamesgorman2 (approved):

jamesgorman2:
Thank you David.

Issue #605, opened 2018-05-02
Writes out of order when using multiple threads


JonasHallFnx:
In #504 an ordering issue for writes was solved by having writes enqueued on the computation scheduler. Unfortunately there seems to be a glitch when the emitted messages comes from the event loop sometimes and from other threads sometimes. When another thread than the event loop is used, the BackpressureManagingHandler starts to enqueue writes, but if the next write happens on the event loop then the write is not enqueued. When this happens, there is a chance that the previous write has not been processed on the computation scheduler yet, which makes the writes out of order.

A possible solution is to continue to enqueue messages until the queue is drained. PR coming up.

PR #606, opened 2018-05-02
Fix writes out of order

JonasHallFnx:
Continue to enqueue writes until queue is drained, to avoid that writes comes out of order when writes are triggered from both the event loop and other threads.

Solves issue #605


JonasHallFnx:
Any maintainers available for review? @jamesgorman2 ?

jamesgorman2 (approved):
LGTM

There might still be a race around the first off event loop write (a write(inEL) concurrent with the _first_ write(!inEL)), but that already exists and isn't introduced by this PR.

@TurboProgramming I'm approving for now, but can you can have think and I'll merge or hold off after I do the same and get feedback from you

JonasHallFnx:
@jamesgorman2 any chance we can get this merged?

PR #607, opened 2018-05-23
PooledConnection.close must be subscribed to be executed

JonasHallFnx:
In RxNetty 0.5.x the `PooledConnection` will just return a `releaseObservable`, as opposed to 0.4.x. The returned Observable must be subscribed in order for the connection to be released. There is a leak of connections in case the subscriber has unsubscribed when the connection is to be emitted.

### DIFF
--- a/rxnetty-common/src/main/java/io/reactivex/netty/client/ClientConnectionToChannelBridge.java
+++ b/rxnetty-common/src/main/java/io/reactivex/netty/client/ClientConnectionToChannelBridge.java
@@ -33,6 +33,8 @@ import io.reactivex.netty.internal.ExecuteInEventloopAction;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import rx.Subscriber;
+import rx.functions.Action1;
+import rx.functions.Actions;
 import rx.observers.SafeSubscriber;
 import rx.subscriptions.Subscriptions;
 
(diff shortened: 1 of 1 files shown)


magnusp (approved):

niklasga (approved):

rxnetty-common/src/main/java/io/reactivex/netty/client/ClientConnectionToChannelBridge.java
```diff
@@ -138,7 +140,13 @@ private void newConnectionReuseEvent(Channel channel, final ConnectionReuseEvent
             subscriber.onNext(event.getPooledConnection());
             checkEagerSubscriptionIfConfigured(channel);
         } else {
-            event.getPooledConnection().close(false); // If pooled connection not sent to the subscriber, release to the pool.
+            // If pooled connection not sent to the subscriber, release to the pool.
+            event.getPooledConnection().close(false).subscribe(Actions.empty(), new Action1<Throwable>() {
```
jepp3, line 144: Can be replaced with lamda style action instead?

jepp3: `
(throwable -> {
			logger.error("Error closing connection.", throwable);
		})`

JonasHallFnx: No can do. Source level is set to 1.7, lambda not allowed.

jepp3 (approved):

jamesgorman2 (approved):
LGTM

jamesgorman2:
@JonasHallFnx this is in v0.5.3-rc1

I'll be able to have another look at #606 this week too, and if it's all good we'll get that into -rc2

Issue #611, opened 2018-07-16
Connection Leak 0.5.2


moderakh:
for our project, https://github.com/Azure/azure-cosmosdb-java, we are trying to upgrade our dependency on RxNetty from 0.4.x to 0.5.2 however it seems there is a connection leak issue in 0.5.2 that we are facing.

code to repro:
```java
public class testing {

    private static class DefaultSSLEngineFactory implements Func1<ByteBufAllocator, SSLEngine> {
        private final SslContext sslContex;

        private DefaultSSLEngineFactory() {
            try {
                SslProvider sslProvider = SslContext.defaultClientProvider();
                sslContex = SslContextBuilder.forClient().sslProvider(sslProvider).build();
            } catch (SSLException e) {
                throw new IllegalStateException("Failed to create default SSL context", e);
            }
        }

        @Override
        public SSLEngine call(ByteBufAllocator byteBufAllocator) {
            return sslContex.newEngine(byteBufAllocator);
        }
    }

    public static void main(String[] args) throws  Exception {

        PoolConfig<ByteBuf, ByteBuf> pConfig = new PoolConfig<>();
        pConfig.maxConnections(1000);
        HttpClient<ByteBuf, ByteBuf> cl = HttpClientImpl.newClient(SingleHostPoolingProviderFactory.create(pConfig),
                Observable.just(new Host(new InetSocketAddress(InetAddress.getByName("github.com"), 443))))
                .secure(new DefaultSSLEngineFactory());

        List<Observable<Void>> list = new ArrayList<>();

        for(int i = 0; i < 4000; i++) {

            Observable<HttpClientResponse<ByteBuf>> rsp =
                    cl.createRequest(HttpMethod.GET, "/").writeBytesContent(Observable.just(new byte[] {}));

            Observable<Void> contentDiscardedObs = rsp.flatMap(hcr -> hcr.discardContent());
            list.add(contentDiscardedObs);
        }

        List<Void> res = Observable.merge(list, 1000)
                .observeOn(Schedulers.computation())
                .toList().toBlocking().single();

        System.out.println("DONE");
    }
}
```

As Connection Pool Size is 1000, and we are merging results with 1000 as the degree of concurrency the above code should work. However we get the following failure:

``` log
Exception in thread "main" java.lang.RuntimeException: io.reactivex.netty.client.pool.PoolExhaustedException: Client connection pool exhausted.
	at rx.exceptions.Exceptions.propagate(Exceptions.java:57)
[...]

Issue #511, opened 2016-06-01
What could cause the RxNetty client to return an empty stream ?


Crystark:
I'm using a fairly simple RxNetty client:

``` java
    public Service(String host, int port, int maxConnection) {
        endpoint = RxNetty
            .<ByteBuf, ByteBuf> newHttpClientBuilder(host, port)
            .withMaxConnections(maxConnection)
            .build();
    }

    public Observable<Response> submit(Observable<Request> o) {
        return o
            .flatMap(br -> endpoint.submit(HttpClientRequest.createPost(br.uri()).withContent(br.bytes())))
            .flatMap(r -> r.getContent()
                .doOnNext(bb -> bb.retain().touch("ByteBuf from proxy response"))
                .switchIfEmpty(Observable.just(null))
                .map(c -> new Response(r.getStatus(), c, r.getHeaders().entries()))
            );
    }
```

The thing is, I often get an empty observable from `endpoint.submit` and i'm not sure what could be causing that. I expect it to always return a `HttpClientResponse<ByteBuf>`.

What may cause it to return an empty stream ?

Note: I checked and my input Observable (`o`) is never empty.

NiteshKant:
> > >  I expect it to always return a HttpClientResponse<ByteBuf>

Yes, `HttpClient.submit()` should either give a response or error, but it should not be empty.
Is this reproducible?

Crystark:
No, I haven't been able to reproduce it yet, only observe it in production. It isn't recent. I've tried switching to previous RxNetty version and still have the same behavior.

Crystark:
I currently fixed it using `switchIfEmpty(DEFAULT_RESPONSE.doOnNext(r -> metricFallback.increment()))` so I can track how often it happens. Visually, we oberved approximatly 5 of those per second for 1500QPS so that seems rare.

Just so you know, I've changed a bit the code to make sure:

``` java
    public Observable<Response> submit(Observable<Request> o) {
        return o
            .switchIfEmpty(Observable.<Request> empty().doOnCompleted(() -> L.warn("BEFORE")))
            .flatMap(br -> endpoint.submit(HttpClientRequest.createPost(br.uri()).withContent(br.bytes()))
                .switchIfEmpty(Observable.<HttpClientResponse<ByteBuf>> empty().doOnCompleted(() -> L.warn("INSIDE"))))
[...]

PR #613, opened 2019-01-14
Fix sometimes empty response.

rassmate:
Sometimes the response will be accidentally empty.
This seems to be due to the fact that onCompleted is sometimes called before onNext where the connection is send. 

Moving the onNext event into the synchronized block fixes the problem.

Also added a test that without is flaky without my fix

Might do some good for
https://github.com/ReactiveX/RxNetty/issues/611
and
https://github.com/ReactiveX/RxNetty/issues/511


jamesgorman2 (approved):
Thanks Jonas, LGTM.

Synchronised sequencing looks right. There's probably a less blocking (and more complicated) approach like in the canonical RxJava examples _but_ I don't think they bring enough benefit to warrant putting them in.

PR #615, opened 2019-01-22
interval instead of timer for the idleConnectionCleanup

rassmate:
Using interval instead of timer for the idleConnectionCleanup because the interval will tick many times every specified millisecond and timer will only tick once and then complete and I guess that is not what we want

### DIFF
--- a/rxnetty-common/src/main/java/io/reactivex/netty/client/pool/PoolConfig.java
+++ b/rxnetty-common/src/main/java/io/reactivex/netty/client/pool/PoolConfig.java
@@ -37,7 +37,7 @@ public class PoolConfig<W, R> {
 
     public PoolConfig() {
         maxIdleTimeMillis = DEFAULT_MAX_IDLE_TIME_MILLIS;
-        idleConnCleanupTicker = Observable.interval(maxIdleTimeMillis, TimeUnit.MILLISECONDS);
+        idleConnCleanupTicker = Observable.interval(maxIdleTimeMillis, maxIdleTimeMillis, TimeUnit.MILLISECONDS);
         idleConnectionsHolder = new FIFOIdleConnectionsHolder<>();
         limitDeterminationStrategy = UnboundedPoolLimitDeterminationStrategy.INSTANCE;
     }
--- a/rxnetty-common/src/main/java/io/reactivex/netty/client/pool/PoolConfig.java
+++ b/rxnetty-common/src/main/java/io/reactivex/netty/client/pool/PoolConfig.java
@@ -37,7 +37,7 @@ public class PoolConfig<W, R> {
 
     public PoolConfig() {
         maxIdleTimeMillis = DEFAULT_MAX_IDLE_TIME_MILLIS;
-        idleConnCleanupTicker = Observable.timer(maxIdleTimeMillis, TimeUnit.MILLISECONDS);
+        idleConnCleanupTicker = Observable.interval(maxIdleTimeMillis, TimeUnit.MILLISECONDS);
         idleConnectionsHolder = new FIFOIdleConnectionsHolder<>();
         limitDeterminationStrategy = UnboundedPoolLimitDeterminationStrategy.INSTANCE;
     }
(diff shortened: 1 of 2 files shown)


rassmate:
And this as well @jamesgorman2

rxnetty-common/src/main/java/io/reactivex/netty/client/pool/PooledConnectionProviderImpl.java
```diff
@@ -70,18 +70,18 @@ public PooledConnectionProviderImpl(PoolConfig<W, R> poolConfig, HostConnector<W
...
-                        logger.error("Ignoring error cleaning up idle connections.",
-                                     throwable);
-                        return Observable.empty();
-                    }
-                }).subscribe(Actions.empty()); // Errors are logged and ignored.
+        idleConnCleanupSubscription = poolConfig.getIdleConnCleanupTicker()
+            .doOnError(LogErrorAction.INSTANCE)
+            .retry() // Retry when there is an error in timer.
+            .concatMap(new IdleConnectionCleanupTask())
+            .onErrorResumeNext(new Func1<Throwable, Observable<Void>>() {
```
jamesgorman2: Looking at this, I think the `onErrorResumeNext(...)` should be replaced by a `doOnError(...)` to log the error, then a `retry()`. 
The current change fixes the problem of the pool only being cleaned up once, but if there is a transient failure then it'll stop cleaning up.

jamesgorman2 (changes requested):
@rassmate could you have a look at my comment and have a think, if I'm wrong we're good to go on this, otherwise can you make the change and we'll get this merged

rxnetty-common/src/main/java/io/reactivex/netty/client/pool/PooledConnectionProviderImpl.java
rassmate: @jamesgorman2 the first commit is obsolete when i found the timer - interval problem but I forgot to revert. Will push revert of that commit.

rassmate: @jamesgorman2 sorry, I now see I did revert back to the original state of the subscription. And the onErrorResumeNext will log the error and then return empty without error so the cleaning up will not stop right? Or how do you mean by transient error?

JonasHallFnx: @rassmate transient probably refers to errors originating from IdleConnectionCleanupTask. An error is a terminal event in Rx so the stream will be aborted. A retry would subscribe the stream again.

rassmate:
Ok, I think I understand.
I’ll fix tomorrow morning

ons 13 feb. 2019 kl. 20:30 skrev JonasHallFnx <notifications@github.com>:

rxnetty-common/src/main/java/io/reactivex/netty/client/pool/PoolConfig.java
```diff
@@ -37,7 +37,7 @@
 
     public PoolConfig() {
         maxIdleTimeMillis = DEFAULT_MAX_IDLE_TIME_MILLIS;
-        idleConnCleanupTicker = Observable.timer(maxIdleTimeMillis, TimeUnit.MILLISECONDS);
+        idleConnCleanupTicker = Observable.interval(maxIdleTimeMillis, TimeUnit.MILLISECONDS);
```
jamesgorman2: Sorry, one more. This should have an initial delay so that it doesn't try to clean up immediately. It's less important on creation, but if there are errors it will keep emitting right after each failure. This means a failure will keep triggering as fast as it can instead of just every `maxIdleTimeMillis`.

```
idleConnCleanupTicker = Observable.interval(maxIdleTimeMillis, maxIdleTimeMillis, TimeUnit.MILLISECONDS);
```

jamesgorman2 (changes requested):
Sorry, something I missed the first time around. Small fix. 
Also just confirmed expected behaviour of the retry in a scratch test.

rassmate:
Good catch, will fix

tors 14 feb. 2019 kl. 10:05 skrev James Gorman <notifications@github.com>:

jamesgorman2 (approved):
LGTM

jamesgorman2:
Thanks Jonas, now rc is in flight.

rassmate:
No problem! Our production servers feels a lot better with these fixes in place.
Thanks for the attention. @jamesgorman2

rassmate:
@jamesgorman2 any non rc release of this anytime soon?

jamesgorman2:
Just pushed out 0.5.3 with the last batch of changes.

PR #620, opened 2020-07-23
Upgrade to Netty 4.1.51-Final to address CVE-2020-11612

berngp:
Versions [4.1.0-Final, 4.1.46.Final) are vulnerable to Uncontrolled Memory Allocation while decoding
a ZlibEncoded byte stream. An attacker could send a large ZlibEncoded byte stream to the Netty server,
forcing the server to allocate all of its free memory to a single decoder.

Ref.
https://snyk.io/vuln/SNYK-JAVA-IONETTY-564897
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-11612

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 language: java
 jdk:
-- oraclejdk8
+  - openjdk8
+# - oraclejdk8
+# addons:
+#   apt:
+#     packages:
+#       - oracle-java8-installer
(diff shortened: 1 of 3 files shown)


berngp:
The PR is using openjdk8 since the build was running into trouble with TravisCI due that it was unable to install `oraclejdk8`.

